PR #60, opened 2017-08-08
Try to vote twice

shaydvir:
Please let me know if this workflow is ok. And the example of try to vote twice is what you meant at:
https://github.com/daostack/daostack/issues/51
Thank you.


jellegerbrandy:
Seems pretty good to me. Thanks, Shay!

Issue #66, opened 2017-08-10
Why does a proposal.owner can vote on behalf of someone else?


assafmo:
https://github.com/daostack/daostack/blob/6a2a928cc130e78638d4f440112e11c147f4b0b3/contracts/VotingMachines/SimpleVote.sol#L106-L108

What is preventing a malicious `proposal.owner` rapidly voting for its proposal "on behalf" of other voters (just call `vote` with other voters' addresses) and triggering a successful `executeProposal(_proposalId)` in https://github.com/daostack/daostack/blob/6a2a928cc130e78638d4f440112e11c147f4b0b3/contracts/VotingMachines/SimpleVote.sol#L133 after enough votes?

I think that `_voter` should always be `msg.sender`. 

Am I missing someting? When and why would a voter **need to** vote on behalf of someone else?

leviadam:
The SimpleVote contract is Voting as a service.
There is no problem of the owner voting for someone because the owner in the usual case is just a scheme.
The idea is to allow the scheme to vote in the name of the agent that proposed the vote.

jellegerbrandy:
I'm just adding these observations:

  * Even if it is intended behavior and so not a  bug, having an all-power proposal owner _is_  a  complication. It means that the  voting procedure is about as secure as the voting contract itself _together with the owner_, which could in theory be any contract (or user). This is not very good practice: we'd like to have a clear division of responsibility to keep our thinking about security sane. 
 * I looked in the code, and the _only_ use I can see for the `owner` argument is to vote automatically for a  proposal that is submitted. This is basically a UX feature and so has no real place on the solidity level, and it could  just as well be a separate transaction.
  * "triggering a successful executeProposal(_proposalId)" -> Again, this could be _any_ executable, so it is hard to say things in general, but the answer here is that the controller will be in charge of security here (de facto saying "I only do stuff that is aksed for by my schemes, and my schemes are responsible for the proposal security")

I'd be leaning towards simplification: taking the owner out, (and thus not having an automatic "yes" vote on proposing).

leviadam:
It is indeed not good to solve UI problems on the blockchain.
Here it is a very annoying UI issue.
I tend to agree we should change it. But the UI cost is quite significant.

assafmo:
Thank you for the quick responses!

I believe that [simplicity follows complexity](http://mitadmissions.org/blogs/entry/the-complexity-of-simpliciyy). The simpler the contract will be, it will have less bugs.

[...]

PR #69, opened 2017-08-13
SimpleVote - 7 more tests (passing) + bugfix for cancelVote

assafmo:


jellegerbrandy:
Really good job, assafmo! Many thanks. Good call on the bug as well.

I'll merge it when I get the time (or perhaps @leviadam  can). 

Given that you have added the _voter arg to cancelVote, we probably also want the "if not specifed, _voter = msg.sender" stanza that is also in vote():

```
        if (msg.sender != proposal.owner) {
            _voter = msg.sender;
        }
```

assafmo:
@jellegerbrandy you are right. I still due think that the owner shouldn't be allow to vote for other parties as we discussed in #66.

assafmo:
@jellegerbrandy I added your remark to the PR.
Those sort of scenarios shoud be **heavily** test.

assafmo:
Added 2 more tests.

leviadam:
Tests are failing. Linting problem.

assafmo:
mmm... What linter should I be using?

leviadam:
eslint.
And you should run `yarn lint` before committing.

PR #80, opened 2017-08-29
Fix broken Reputation tests

assafmo:


leviadam:
Hi ;)
The changes here will work fine.
But I think we should stick to better practices.
There is a practice how to catch an error from the EVM. You can see it in the absolutevote.js file.

assafmo:
Oh, I see. Okay, I'll fix this.

leviadam:
Great! Thanks!

PR #88, opened 2017-10-17
Package DAOStack for npm and application reuse

dkent600:
- Enable packaging for reuse via NPM (see https://www.npmjs.com/package/emergent-arc)
- Wrap `require` statements on .sol files so that they work in the context of both testing/migration and applications.
- Add `arc` module as the central starting point for application reuse.

See [daostack-frontend-aurelia](https://github.com/daostack/daostack-frontend-aurelia) for example of reuse using NPM package (at this writing, is in the branch "useEmergentArcPackage").


dkent600:
@leviadam Heads up: There are a couple of lines in `settings.js` that are questionable to me.  Why are we grabbing the js instead of calling requireContract?

To be consistent, should we not be doing the same thing with `SimpleVote` and `SchemeRegistrar`?

```
//const GlobalConstraintRegistrar = requireContract("GlobalConstraintRegistrar"); 
import { GlobalConstraintRegistrar } from   '../lib/globalconstraintregistrar.js';

// const SimpleContributionScheme = requireContract("SimpleContributionScheme");
import { SimpleContributionScheme } from   '../lib/simplecontributionscheme.js';
```

Issue #95, opened 2017-10-30
Migration to kovan fails on third stage


dkent600:
Migrating to kovan fails with the following error:

```
Running migration: 3_deploy_DAOstack_SAFT.js
Error encountered, bailing. Network state unknown. Review successful transactions manually.
TypeError: Cannot convert undefined or null to object
    at C:\Users\dkent\AppData\Roaming\npm\node_modules\truffle\build\cli.bundled.js:96299:16
```

dkent600:
In 3_deploy_DAOstack_SAFT.js, `wallet = web3.eth.accounts[3]` needs to be `wallet = web3.eth.accounts[0]`

PR #96, opened 2017-10-30
#95 -- Migration to kovan fails on third stage 

dkent600:
Fix for #95 -- Migration to kovan fails on third stage


dkent600:
This was created against master.  Should've been against dev.

Issue #104, opened 2017-11-05
Arc should provide sets of available schemes and voting machines


dkent600:
Currently Arc just provides a list of contracts.  Applications need to have a way to obtain distinct lists of the available schemes and available voting machines.

PR #105, opened 2017-11-05
#104 Arc should provide sets of available schemes and voting machines 

dkent600:
#104 Arc should provide sets of available schemes and voting machines


leviadam:
We are not pulling to master, only to dev.

dkent600:
Dang, sorry, I tend to forget to select dev in the dropdown

Issue #112, opened 2017-11-13
calls to `assert` are crashing application code


dkent600:
Calls to `assert` are crashing application code.  They need to be removed or stubbed-out.

leviadam:
Do we have any idea why is that happening ?

leviadam:
OK, we decided to remove the `asset` for the time being.

PR #113, opened 2017-11-13
Commented-out calls to assert

dkent600:
#112 - calls to `assert` are crashing application code


leviadam:
Not sure about deleting the `asset` commands.
Is there a reason why it crashes things?

dkent600:
@leviadam It crashes because `assert` doesn't exist outside of the test environment

dkent600:
The alternative is to stub it out when it doesn't exist.  I can do that if you prefer.

leviadam:
No, we can go without asserts in the lib, and have tests.

Issue #119, opened 2017-11-14
default user's gas limit is too high


dkent600:
The default user's gas limit is too high.  As a result, the Alchemy user must enter a working value on approving every transaction against Kovan.

The value that is working for me is  4465000.

The value is configured in utils.js.

leviadam:
The problem is that for some TXs we need more than that.

dkent600:
@leviadam All I know is that this setting is working a heck of a lot better than the previous one.  I haven't yet encountered a problem with it.  Do you know what transactions are potentially a problem, and how much gas they will need?

PR #120, opened 2017-11-14
#119 -- default user's gas limit is too high

dkent600:
#119 
Set to 465000


leviadam:
Too high for which chain? Kovan?

dkent600:
Kovan

Issue #121, opened 2017-11-14
Alchemy can't find arc.d.ts


dkent600:
Alchemy can't find arc.d.ts and is failing to compile.  I was mistaken when I thought earlier that it was finding it and working OK, but I was unknowingly testing against the (previous) v8 of the library.

dkent600:
Nevermind, turns out it was the module name in the d.ts.  Was already fixed by a direct commit to dev by @tibetsprague

PR #122, opened 2017-11-14
#121Alchemy can't find arc.d.ts 

dkent600:
#121 -- Alchemy can't find arc.d.ts 

I fixed this by moving `arc.d.ts` into the root folder of the project.  I realize that this pulls it out of the lib folder and maybe that is not desired.  But the way npm works, I don't know of an alternative short of creating a separate package structure in the lib folder and restructuring things thereunder.


dkent600:
forgot to set against dev instead of master

PR #123, opened 2017-11-14
#121Alchemy can't find arc.d.ts 

dkent600:
#121 -- Alchemy can't find arc.d.ts 

I fixed this by moving `arc.d.ts` into the root folder of the project.  I realize that this pulls it out of the lib folder and maybe that is not desired.  But the way npm works, I don't know of an alternative short of creating a separate package structure in the lib folder and restructuring things thereunder.


dkent600:
I'm going to have to look at this some more.  I was fooled again into thinking it was working, but it still isn't.  I don't know what is the problem, but my #1 theory at the moment is that there is something wrong with the d.ts itself.  I'll post here when I know more...

tibetsprague:
@dkent600 its probably because you changed the name of the package but not the module in the definitions file

tibetsprague:
i actually just pushed that fix to dev

dkent600:
@tibetsprague Bingo.  I can back out moving the d.ts.  The problem was not with finding the file, it was that the file was declaring the module "emergent-arc" instead of "daostack-arc".  Doh.  I'll update the PR accordingly.

dkent600:
@tibetsprague  I didn't see your prior comments on this above.  Didn't know you made the change to dev.  I'll just close this PR.

PR #137, opened 2017-11-22
Get wallet stuff fully usable by a third party

tibetsprague:
Finish up getEtherBalance() and getOrgTokenBalance() functions
Add the compiled contracts to the repo as they are needed by Arc.js
Fix up typescript definitions


.gitignore
```diff
@@ -1,6 +1,5 @@
 *sublime*
 *.pyc
-build/
```
leviadam: Not sure this is the way to do things.
But we give it a try.

Issue #152, opened 2017-12-06
Bug: UpgradeScheme.proposeChangeUpgradingScheme doesn't change the upgrading scheme


dkent600:
`UpgradeScheme.proposeChangeUpgradingScheme` does not explicitely fail, but it doesn't appear to work either.  The following test code, commented-out because it fails, is in branch #98, demonstrating the repro case.  It generates the errors:

"AssertionError: new scheme is not registered into the controller: expected false to be true"

and

"AssertionError: original scheme is still registered into the controller: expected true to be false"

```
  it("proposeUpgradingScheme javascript wrapper should change upgrade scheme", async function() {
    const organization = await forgeOrganization();

    let upgradeScheme = await organization.scheme('UpgradeScheme');
    const newUpgradeScheme = await UpgradeScheme.new(SOME_ADDRESS, 0, SOME_ADDRESS);

    assert.isFalse(await organization.controller.isSchemeRegistered(newUpgradeScheme.address), "new scheme is already registered into the controller");
    assert.isTrue(await organization.controller.isSchemeRegistered(upgradeScheme.address), "original scheme is not registered into the controller");

    let tx = await upgradeScheme.proposeUpgradingScheme({
      avatar: organization.avatar.address,
      scheme: newUpgradeScheme.address,
      schemeParametersHash: await organization.controller.getSchemeParameters(newUpgradeScheme.address)
    });

    // newUpgradeScheme.registerOrganization(organization.avatar.address);

    const proposalId = getValueFromLogs(tx, '_proposalId');

    organization.vote(proposalId, 1, {from: accounts[2]});

    assert.isTrue(await organization.controller.isSchemeRegistered(newUpgradeScheme.address), "new scheme is not registered into the controller");

    assert.isFalse(await organization.controller.isSchemeRegistered(upgradeScheme.address), "original scheme is still registered into the controller");
  });
```

leviadam:
Fixed it. It was a permission problem in the library.
Making a pull request now.

Issue #151, opened 2017-12-06
Bug: organization.schemes does not reflect current state of scheme registration


dkent600:
`organization.schemes` has a couple of problems:

1) when a given scheme has been modified then `organization.schemes` returns two identical records of the modified scheme.  Three modifications then three records, and so on.

The fix could be that `SchemeRegistrar` fire different events when modifying versus when adding a scheme. Barring that, `organization.schemes` will need to check for dups.

2)  when a given scheme has been unregistered, then `organization.schemes` still returns a record for the removed scheme.

`organization.schemes` should be checking UnregisterScheme events.

Note:  I have code in Alchemy that already does all of this correctly.

leviadam:
I think this better be handled off chain.

dkent600:
k, i'll copy or move my code from Aurelia into organizations.

dkent600:
Some code was merged to address this.  There are an improvement, but not 100% sufficient.  Final changes to come in a separate PR.

dkent600:
The code for this has been moved over to the arc-js repo.  So this daostack issue can be closed.

Issue #98, opened 2017-10-31
Organization.proposeScheme should be completed or removed


dkent600:
`Organization.proposeScheme` only supports a couple of schemes.  It should be completed (and refactored) or removed until it is complete.

leviadam:
We have to refactor the whole thing. This is part of redesigning the package.
After we will work on the settings this will be the next step with @tibetsprague.

dkent600:
I have written some code in Alchemy that could be moved over to Arc in this effort, though even at that, there is still a lot to do.

dkent600:
I'm going to use this issue as a placeholder for creating javascript helper methods for all of the primary scheme functions.

dkent600:
Fixed by #153

PR #153, opened 2017-12-06
#98 Create javascript wrappers for all the contract functionality

dkent600:
This adds javascript wrappers for all the contract functionality.  It also adds lots of declarations in arc.d,ts.

Still not done, but still could be if desired, or else could be done in separate PRs:

1) regarding TokenCapGC and AbsoluteVote:  I've added wrappers for them, but no contract functionality.  AbsoluteVote seems to be in a lot of flux, so maybe best to hold off?  I could do TokenCapGC though.

2) there are a couple of tests that are commented-out because they fail.  I've created issues for them:  #151 and #152.


lib/globalconstraintregistrar.js
```diff
@@ -30,6 +30,91 @@ export class GlobalConstraintRegistrar extends ExtendTruffleContract(SolidityGlo
     return new this(contract);
   }
 
+  async proposeToAddModifyGlobalConstraint(opts={}) {
+    const defaults = {
```
tibetsprague: why are we using this model of a config object instead of parameters to the function if all the options are going to be undefined and throw an error if they don't exist? did we decide at some point that we wanted to use options instead of specific parameters for all functions in arc?

lib/organization.js
```diff
@@ -278,174 +269,174 @@ export class Organization {
...
-    // calculate (and set) the hash that will be used to remove the parameters
-    await AbsoluteVote.at(options.votingMachine).setParameters(options.reputation, options.absPrecReq, true);
-    options.votingMachineHash = await AbsoluteVote.at(options.votingMachine).getParametersHash(options.reputation, options.absPrecReq, true);
-
-    const globalConstraintRegistrar = await this.scheme('GlobalConstraintRegistrar');
-    let tx = await globalConstraintRegistrar.proposeGlobalConstraint(this.avatar.address, options.address, options.paramsHash, options.votingMachineHash);
-    const proposalId = getValueFromLogs(tx, '_proposalId');
-    return proposalId;
-  }
+  // async proposeScheme(opts={}) {
```
tibetsprague: why is all this commented out for now?

lib/simplecontributionscheme.js
```diff
@@ -31,20 +31,86 @@ class SimpleContributionScheme extends ExtendTruffleContract(SoliditySimpleContr
     return new this(contract);
   }
 
-  async submitContribution(opts={}) {
+  async proposeContribution(opts={}) {
```
tibetsprague: why proposeContribution? the contract calls it submitContribution

lib/upgradescheme.js
```diff
@@ -30,6 +31,149 @@ export class UpgradeScheme extends ExtendTruffleContract(SolidityUpgradeScheme)
...
+    {
+      throw new Error("avatar address is not defined");
+    }
+
+    if (!options.controller)
+    {
+      throw new Error("controller address is not defined");
+    }
+
+      let tx = await this.contract.proposeUpgrade(options.avatar, options.controller);
```
tibetsprague: wrong indentation

lib/upgradescheme.js
```diff
@@ -30,6 +31,149 @@ export class UpgradeScheme extends ExtendTruffleContract(SolidityUpgradeScheme)
...
+/********************************************
+ * proposeUpgradingScheme
+ */
+    async proposeUpgradingScheme(opts={}) {
+
+    /**
+     * Note that explicitly supplying any property with a value of undefined will prevent the property
+     * from taking on its default value (weird behavior of default-options)
+     */
+    const defaults = {
```
tibetsprague: indentation off throughout this function

lib/upgradescheme.js
```diff
@@ -30,6 +31,149 @@ export class UpgradeScheme extends ExtendTruffleContract(SolidityUpgradeScheme)
...
+       * obtain the values directly from the submitted scheme.
+       * Otherwise tokenAddress is required.
+       */
+      , tokenAddress: null
+    };
+
+    const options = dopts(opts, defaults, { allowUnknown: true });
+
+    if (!options.avatar)
+    {
```
tibetsprague: we should choose a style of where to put the {, you go back and forth in this code. I prefer putting it on the same line as the conditional

lib/globalconstraintregistrar.js
dkent600: I was continuing with the pre-existing pattern.  But I do prefer to use an options object like that.  I find it tends to be easier to make changes to an API while maintaining backwards compatibility.

lib/organization.js
dkent600: I should have deleted those comments.  I plan to do that.

lib/simplecontributionscheme.js
dkent600: I am aiming for consistency in function naming.  As a result there are more than one whose names differ from the primary underlying contract function.  Further, when a proposal is involved, I want the function name to be explicit that we are proposing something, and again to be consistent in that nomenclature.  I don't see a lot of advantage in matching a contract function name.

lib/upgradescheme.js
dkent600: Would be nice if lint would enforce the stuff we care about.

tibetsprague: definitely

Issue #140, opened 2017-11-26
Bug: UpgradingScheme removes itself when its own address is passed as new UpgradingScheme


dkent600:
UpgradingScheme removes itself when its own address is passed as the new UpgradingScheme in `proposeChangeUpgradingScheme`.

The result is that the user ends up with a corrupt DAO.

To handle this case `proposeChangeUpgradingScheme` should revert.

leviadam:
Great work. Fixed.

dkent600:
I am still ending up with a corrupt DAO in this scenario.

leviadam:
How come? I updated the code so that it will not remove itself.

dkent600:
In `organization.at`, the following line:

`var reputationAddress = await org.controller.nativeReputation();`

throws the following exception:

"Error: VM Exception while executing eth_call: invalid opcode"

leviadam:
This is very weird, reputation have nothing to do with it.
Are you sure you just tried to change the upgrading scheme?
Maybe you actually performed an upgrade and replaced the controller?

dkent600:
I'm definitely not changing the controller.  I confirmed that the crash is happening in `nativeReputation`

leviadam:
Found it. It was a bug in the backend! Good job!
Fixed it.

dkent600:
Now I find that in this case `proposeChangeUpgradingScheme` is throwing an exception, "revert"

leviadam:
This was a permission issue, solved it.

dkent600:
If this is fixed we should add a test for it:

```
it("proposeUpgradingScheme javascript wrapper should modify the modifying scheme", async function() {

    const organization = await forgeOrganization();

    let upgradeScheme = await organization.scheme('UpgradeScheme');

    assert.isTrue(await organization.controller.isSchemeRegistered(upgradeScheme.address), "upgrade scheme is not registered into the controller");

    let tx = await upgradeScheme.proposeUpgradingScheme({
      avatar: organization.avatar.address,
      scheme: upgradeScheme.address,
      schemeParametersHash: SOME_HASH
    });

    // newUpgradeScheme.registerOrganization(organization.avatar.address);

    const proposalId = getValueFromLogs(tx, '_proposalId');

    organization.vote(proposalId, 1, {from: accounts[2]});

    assert.isTrue(await organization.controller.isSchemeRegistered(upgradeScheme.address), "upgrade scheme is no longer registered into the controller");

[...]

PR #159, opened 2017-12-08
Fixing default permissions in the library

leviadam:
This PR fixes the default permissions in the library. Solving issue #152 .
A test is added to check the change of upgrading scheme.


dkent600:
This is also directed at solving #140.  We need to add an automated test for that case as well.

PR #160, opened 2017-12-10
#151 organization.schemes does not reflect current state of scheme registration

dkent600:
#151: organization.schemes does not reflect current state of scheme registration
Also added an automated test for #140 UpgradingScheme removes itself when its own address is passed as new UpgradingScheme


dkent600:
accidently did this against master instead of dev

PR #161, opened 2017-12-10
#151: organization.schemes does not reflect current state of schemes...

dkent600:
#151: organization.schemes does not reflect current state of scheme registration

Improvements to type definitions (arc.d.ts)

Added an automated test for #140 UpgradingScheme removes itself when its own address is passed as new UpgradingScheme

### DIFF
--- a/dist/arc.js
+++ b/dist/arc.js
@@ -1,115 +1,115 @@
 'use strict';
 
 Object.defineProperty(exports, "__esModule", {
-  value: true
+    value: true
 });
 
 var _organization = require('./organization.js');
 
 Object.keys(_organization).forEach(function (key) {
-  if (key === "default" || key === "__esModule") return;
-  Object.defineProperty(exports, key, {
-    enumerable: true,
-    get: function get() {
-      return _organization[key];
-    }
-  });
+    if (key === "default" || key === "__esModule") return;
+    Object.defineProperty(exports, key, {
+        enumerable: true,
+        get: function get() {
+            return _organization[key];
+        }
+    });
 });
 
 var _globalconstraintregistrar = require('./globalconstraintregistrar.js');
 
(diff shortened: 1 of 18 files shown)


leviadam:
Hi @dkent600,

I'm not sure about using the wrappers in the tests.
Rethinking, maybe to take the library to a different repo.
What do you think @orenyodfat @tibetsprague ?

orenyodfat:
I support for having two separate repositories  : 
1. Solidity contracts + contracts tests.
2. JS library  + tests.

I guess it will be easier to maintain and stabilize versions of daostack stack's components.

leviadam:
OK, so it is decided. We go with two repos. So I close this PR. Move things to the new repo. And @dkent600 will PR to that repo.

dist/arc.js
```diff
@@ -1,115 +1,115 @@
 'use strict';
 
 Object.defineProperty(exports, "__esModule", {
-  value: true
+    value: true
```
tibetsprague: i prefer 2 spaces myself

leviadam: LOL. We just moved from 2 to 4 in the solidity :)

dkent600: hmm, that is I set lint to -- 2.

dkent600: @tibetsprague You should look at the code in the PR in arc-js, not here.

dkent600: This PR is closed and the code changes obsolete, at least with respect to lint-related improvements.

dkent600: @tibetsprague  Do you recommend 4 then, or 2?  I don't have a strong preference.

tibetsprague: I much prefer 2 spaces. but im also ok with whatever

PR #171, opened 2017-12-13
Removing all library tests

leviadam:
PR:
1. Removes the build library.
2. Remove library tests.
3. Add check before paying a scheme that the fee is nonzero.


orenyodfat:
@leviadam This pr is against master . Is that the intention ?

leviadam:
No. It is a mistake.

PR #179, opened 2017-12-18
Fix simpleico contract and add tests

orenyodfat:
-Fix issues at simpleico.sol : 
1. "donate" function did not store correctly the totalEthRaised.
2. "setParameters" .
3. no need for fallback function any more.
 "Contracts that receive Ether directly (without a function call, i.e. using send or transfer) but do not define a fallback function throw an exception, sending back the Ether (this was different before Solidity v0.4.0). So if you want your contract to receive Ether, you have to implement a fallback function."
http://solidity.readthedocs.io/en/develop/contracts.html#fallback-function

-Add tests.


orenyodfat:
addressed by #178

PR #183, opened 2017-12-20
Fix Slack Link

Alonski:


Alonski:
BTW, maybe the slack link should link to a service that sends out Slack invitations?
Something like: https://github.com/rauchg/slackin
I can get this setup if I had the resources

leviadam:
I think we have to decide if slack is our main channel. After we decide it might be worth it.
Thanks.

Alonski:
@leviadam Is there a current chat for the project, not necessarily slack?

leviadam:
At the moment we still use slack.
https://daostack.slack.com

Alonski:
@leviadam Is it possible to get an invite so I can see where I can contribute more?

PR #198, opened 2017-12-27
added test/tokencapgc.js {garbage test}

dev-matan-tsuberi:


test/tokencapgc.js
```diff
@@ -45,4 +45,18 @@ contract('TokenCapGC', function (accounts)  {
...
   });
+
+  it('garbage transaction test', async () => {
+    const tokenCapGC = await TokenCapGC.new();
+    const addr = tokenCapGC.address;
+
+    // transaction should fail.
+    try{
+      await web3.eth.call({to: addr, data: 0x231ef231ef231ef});
+      assert.fail(true,false);
```
orenyodfat: It worth throwing some meaningful message here..
Take a look [here](https://github.com/daostack/daostack/blob/master/test/schemeregistrar.js#L77) for reference

dev-matan-tsuberi: Thanks, will do.

orenyodfat:
Why this test is needed ? It is general solidity behavior..calling a contract without a fallback func..
Most of our contacts does not have fallbacks.

dev-matan-tsuberi:
I agree, it's a trivial test. 
I'm new here so I'm just making sure I have everything set up correctly.

orenyodfat:
👍

PR #202, opened 2017-12-27
Refactored Avatar's public interface into a separate interface contract + wrote a simple mock implementing this interface for better testability.

dev-matan-tsuberi:
Again, I'm new here. So feel free to shout at me for possibly making a total mess.

As I was writing tests it occurred to me that we should decouple our implementation from our public interface for greater test-ability. We do not need to instantiate all the contracts on every test, only simple mocks of the direct dependencies.

Note: it still compiles and all tests pass.


test/tokencapgc.js
```diff
@@ -45,4 +46,18 @@ contract('TokenCapGC', function (accounts)  {
     //token total supply is 101
     assert.equal(post,true);
   });
+
+  it('garbage transaction test', async () => {
```
orenyodfat: Please remove this

contracts/test/AvatarMock.sol
```diff
@@ -0,0 +1,34 @@
...
+            _value
+        );
+    }
+    function externalTokenIncreaseApproval(StandardToken _externalToken, address _spender, uint _addedValue) public onlyOwner returns(bool) {
+        ExternalTokenIncreaseApproval(_externalToken, _spender, _addedValue);
+    }
+    function externalTokenDecreaseApproval(StandardToken _externalToken, address _spender, uint _subtractedValue) public onlyOwner returns(bool){
+        ExternalTokenDecreaseApproval(_externalToken, _spender, _subtractedValue);
+    }
+}
```
orenyodfat: solidity linter issue..
check using  solium

orenyodfat:
@dev-matan-tsuberi not sure it will be good practice to add more code (interface) just for easiness of the testing process.

Also , in general I guess it will be good practice to add a test for each new solidity function/file.

dev-matan-tsuberi:
Why not? Think of the following example:
I want to test the simple function `UniversalScheme.registerOrganization(Avatar avatar)`.

Option 1: 
I need to create an `Avatar` contract satisfy all of it's dependencies, which includes a `Reputation` contract. So I need to create a `Reputation` contract, satisfy all of it's dependencies, and so on...
So finally I can test my simple function, and possibly get an irrelevant error which is related to a totally different method on a totally different contract. 

Option 2:
Create a zero-dependency mock version of `Avatar` which simply logs all of it's method calls.
That way we can make sure `registerOrganization` calls the relevent `Avatar` methods. While testing only the method we are interested in.

Other benefits:
1. From the implementation side, it provides modularity; we can create different implementations of `AvatarInterface`.
2. It provides good documentation of how we should use this contract externally.

orenyodfat:
> From the implementation side, it provides modularity; we can create different implementations of AvatarInterface.
Interface is needed to make sure multiple contracts implement the same interface .For Avatar it is not the case as there is only single avatar contract in the system. 

> It provides good documentation of how we should use this contract externally.
We should write a document for each contract which explain the contracts apis and usage.

dev-matan-tsuberi:
Ok, I'll think of something better.

dkent600:
Hmm, my 2c:  I am not sure why it is necessarily not a good practice "to add more code (interface) just for easiness of the testing process".  If such a change truly facilitates testing, and it sounds like it would, thus leading to better quality code and more efficient programming in the long run, then that is a powerful benefit.  The question, in my mind, is more about how much complexity or risk is introduced to the code itself by the addition of the interface -- this is the potential offsetting cost.  Sometimes making code more testable can actually  _reduce_ the inherent risk in the code.  I don't know if that is the case in this proposal, but I think we can be clear about it.

@orenyodfat @dev-matan-tsuberi

orenyodfat:
Thanks @dkent600 for clarification.
In this case I am not sure how it will reduce the risk .

Saying that , we might need interfaces for other future reasons  (e.g upgrade contracts) ....not sure.

PR #207, opened 2017-12-31
Testrpc is change to ganache-cli which is its new name

leviadam:


orenyodfat:
Need to update travis aswell ?

Issue #209, opened 2018-01-02
can't migrate contracts without zeppelin-solidity


dkent600:
We need to include zeppelin-solidity in the arc package dependencies so that contracts can be migrated from the NPM package.

This is following from https://github.com/daostack/arc-js/issues/17, which will require that this change be published to NPM before it can be merged.

I have created a branch (locally) with this change, but am currently unable to commit it due to git authentication issues.

dkent600:
Actually, I think there is a cleaner solution here:  *All* of the dependencies can be moved into devDependencies.  None of them are actually needed in the package.

Arc-js will grab zeppelin before it migrates.  That is sufficient.

dkent600:
Also need to include unmigrated .json files in the package.

PR #217, opened 2018-01-03
Update npm

orenyodfat:
add .npmignore to add contracts jsons to the npm. see issue #209 
update .gitignore.

### DIFF
--- a/.npmignore
+++ b/.npmignore
@@ -1,7 +1,4 @@
 *sublime*
-*.pyc
-build/
-node_modules/
 .node-xmlhttprequest-sync*
 *.tgz
 .vscode/settings.json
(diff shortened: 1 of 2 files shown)


orenyodfat:
This PR just add .npmignore to include the build dir in the npm.
How it related to the dependencies ?

dkent600:
There is no need to add an .npmignore file.  What I've done is simpler and also deals with dependency changes.

dkent600:
BTW, what I've done is safer too.  Instead of specifying what should *not* go in, it specifies what *should* go in.  That is safer, helps prevent unwanted stuff from getting into the package.

orenyodfat:
I added .npmignore inorder to include the build dir which is excluded by .gitignore 
https://docs.npmjs.com/cli/publish#description

dkent600:
I did the same thing by adding it to package.json.

orenyodfat:
I am closing this PR as @dkent600 already done work in this direction.

PR #219, opened 2018-01-03
Changes for end-to-end scripting, integration

dkent600:
Addresses #209 

Changes to package.json:

Moves all dependencies from dependencies into devDependencies.  This is better for consuming apps like arc-js.  When arc-js, for example, installs arc, those spurious arc dependencies won't have to also be installed, and no-one going forward will have to wonder why those dependencies are there.

Added  "files" entry to control what is being packaged, and to ensure that the contract json files are published (overriding  that they are excluded in .gitignore).

Bumped up the package version #.

https://github.com/daostack/arc-js/issues/17 depends on this change being published to npm, so if we could do so soon it would be great!


.gitignore
```diff
@@ -11,6 +11,4 @@ installed_contracts/
 remarks.md
```
orenyodfat: can remarks.md be removed ?

package.json
```diff
@@ -1,24 +1,36 @@
...
   "name": "daostack-arc",
-  "version": "0.0.0-alpha.19",
+  "version": "0.0.0-alpha.25",
   "description": "A platform for building DAOs",
+  "files": [
+    "contracts/",
+    "docs/",
+    "build/contracts/",
+    "migrations/",
+    "truffle.js"
```
orenyodfat: please add test/

dkent600: I'm curious why "test" needs to be in the package?

.gitignore
dkent600: done

package.json
orenyodfat: how about solium and eslinter config files?

dkent600: I'm not clear why any of that needs to be in there.  This is for people who are merely including the package (like arc-js does).  Packages aren't used for development and testing.

Could be I'm missing something....that is just in my experience.

orenyodfat: if images/ is not in the npm so it can be totally removed from this repo.

orenyodfat: > Packages aren't used for development and testing.

I guess that if we include the .sol files we should include tests + linter files.

.gitignore
```diff
@@ -8,9 +8,6 @@ NOTES.md
 test/play.js
```
orenyodfat: remove  : 
 test/play.js
tmp/
installed_contracts/

why we need that here ?

package.json
dkent600: > if images/ is not in the npm so it can be totally removed from this repo.

do you want that included in this PR?

.gitignore
dkent600: I don't know, it wasn't in the scope of this

package.json
orenyodfat: why not ?

orenyodfat: Please add also tsconfig.json

.gitignore
dkent600: but I don't see a reason for them other than maybe other devs want to keep them out

orenyodfat: I think it is historical remaining .lets remove that.

package.json
dkent600: > why not ?

some would complain that the requested changes aren't part of the scope of this issue.  but i'm certainly that last person to so complain...  ;-)

dkent600: But I _would_ still like to understand why you are asking for the inclusion of the various entries.

.gitignore
dkent600: done

package.json
orenyodfat: because if we are publishing code (solidity files) and its binary we also should publish tests ,compiler configuration  and env setup for these files .so any one who would like to make use of that or rebuild that can use the same setup as daostack.

dkent600: > if images/ is not in the npm so it can be totally removed from this repo.

done

dkent600: >because if we are publishing code (solidity files) and its binary we also should publish tests ,compiler configuration and env setup for these files .so any one who would like to make use of that or rebuild that can use the same setup as daostack.

In that case I believe they would obtain the repo from git, they wouldn't use an npm package.

.gitignore
```diff
@@ -5,9 +5,6 @@ logs/
 fabfile.py
```
orenyodfat: fabfile.py ?
NOTES.md ?

package.json
dkent600: Corollary:   We probably should define a release in git every time we publish to NPM.

.gitignore
dkent600: no idea.  do you have anymore  to request while i'm in that file :-)

package.json
orenyodfat: We can enable that not just via git. easy.

.gitignore
orenyodfat: logs/ ?

package.json
dkent600: cool

.gitignore
dkent600: hmm, i wonder if logs might be created in a diagnosing mode of any of our tools.

package.json
orenyodfat: So to support this option need to get back test/ and all other env config files.
I guess it will be easy and cleaner to remove "file" in package.json and add .npmignore.

dkent600: that's weird, and not worth it if you ask me.

orenyodfat: This "So to support this option need to get back test/ and all other env config files." ?
or
This "I guess it will be easy and cleaner to remove "file" in package.json and add .npmignore but not so important." 

or both ?

dkent600: lol

this:

> So to support this option need to get back test/ and all other env config files.

I'm not in favor of having to do that unless there is a good reason.

orenyodfat: please add also contribution.md.

dkent600: 'contribution.md' is also not relevant to a package, is it?  Only to the repo...

orenyodfat: > So to support this option need to get back test/ and all other env config files.

I am not have a very strong opinion here..but i have a feeling that if we publish code we should publish tests as well.
@leviadam what your opinion re that ?

dkent600: My position:  In my experience, packages aren't used for development and testing; they are kept minimal, just what is necessary to execute, not to test and develop.  Tests come with the repo.

orenyodfat: the necessary to execute (in our case deploy) are the contracts binaries.
Though this npm might be used also for dev.
I would even say that it will be nice to enable 'npm test' for this package by adding  script/test.sh which build and deploy the contracts on testrpc and run the tests on it.

dkent600: Yes, such a test script could be useful.  Just to be clear, fixing it in package.json does not imply that the test environment files need to be included in the NPM package.

orenyodfat: at least you need the tests ,tsconfig.json .no ?

dkent600: I can't think of any reason why we would support development via the package, nor why anyone would want to or need to.  I also can't think of any other package that does such a thing (though of course I am only familiar with a tiny percentage of npm packages).

I'm open to suggestions though...

dkent600: > at least you need the tests ,tsconfig.json .no ?

One runs tests from the repo, not the package.

orenyodfat: You are familiar with https://github.com/OpenZeppelin/zeppelin-solidity .right ?

dkent600: If someone ever asks for it, or we can anticipate right now why someone would be likely to need it, then sure, let's include the additional test-related stuff in the package.   Otherwise I don't recommend it.

orenyodfat: You know this one  i guess https://www.npmjs.com/package/zeppelin-solidity

dkent600: Indeed they do.  Any idea why?

I can find lots of counter examples.

orenyodfat: If you include some code (library or any other) to your product (via npm or what ever) you would like to have the ability to test it and to be sure that you are testing and using the right thing.

dkent600: >  f you include some code (library or any other) to your product (via npm or what ever) you would like to have the ability to test it and to be sure that you are testing and using the right thing.

And then we have to support them as they try to get the tools to work....

I don't know anyone who does that, and one can see that the tests passed on NPM or github.

It looks to me like most of the modules being pulled into arc right now don't include test code.

dkent600: Just to be clear:  If you are convinced this is a need and want to make it happen, I support you.  I'm just giving my position, open to change, until a decision is made.

orenyodfat:
commit 656f6b4 fix issue #220

package.json
```diff
@@ -1,24 +1,36 @@
...
+    "build/contracts/",
+    "migrations/",
+    "test/",
+    "truffle.js",
+    "tsconfig.json"
+  ],
   "scripts": {
-    "test": "truffle test",
-    "ganache": "ganache-cli -l 6500000 --account=\"0x0191ecbd1b150b8a3c27c27010ba51b45521689611e669109e034fd66ae69621,9999999999999999999999999999999999999999999\" --account=\"0x00f360233e89c65970a41d4a85990ec6669526b2230e867c352130151453180d,9999999999999999999999999999999999999999999\" --account=\"0x987a26abca7432016104ce2f24ce639340e25afe06ac69f68791399e7a5d1028,9999999999999999999999999999999999999999999\" --account=\"0x89af34b1b7347834048b99423dad174a14bf14540d720d72c16ae92e94b987cb,9999999999999999999999999999999999999999999\" --account=\"0xc867be647eb2bc51e4c0d61066859875cf3634fe949b6f5f85c69ab90e485b37,9999999999999999999999999999999999999999999\" --account=\"0xefabcc2377dee5e51b5a9e65a3854aec85fbbec3cb584d8ad4f9679869fb33c6,9999999999999999999999999999999999999999999\"",
+    "test": "./node_modules/.bin/run-with-testrpc --testrpc-cmd ganache-cli --gasLimit 6500000 'truffle compile && truffle migrate && truffle test'",
```
leviadam: I think we still need a script for ganache. We can have a single script for test, but still.
I'm not sure we need a truffle compilt & migrate & test.
I think `truffle test` is enough.

orenyodfat: why need a script for ganache ?
The reason for  'compilt & migrate & test' and not 'test' is because I  found it more descriptive of what the test is actually doing.

package.json
```diff
@@ -27,14 +39,21 @@
     "eslint-plugin-promise": "^3.6.0",
     "eslint-plugin-react": "^7.5.1",
     "eslint-plugin-standard": "^3.0.1",
+    "ethers": "^2.1.3",
```
tibetsprague: we dont need this anymore, this is for arc-js

package.json
```diff
@@ -1,24 +1,36 @@
...
   "devDependencies": {
+    "@types/bignumber.js": "^4.0.3",
     "babel-cli": "^6.26.0",
     "babel-eslint": "^8.1.2",
     "babel-plugin-syntax-async-functions": "^6.13.0",
     "babel-polyfill": "^6.26.0",
     "babel-preset-es2015": "^6.24.1",
     "babel-register": "^6.26.0",
     "bignumber.js": "^5.0.0",
+    "bip39": "^2.4.0",
```
tibetsprague: we dont need this anymore, this is for arc-js

package.json
```diff
@@ -1,24 +1,36 @@
...
     "start": "pm2 start truffle -- serve",
     "lint": "eslint .",
     "lint --fix": "eslint --fix .",
     "solium": "solium --dir ./contracts/",
-    "truffle": "truffle"
+    "truffle": "truffle",
+    "build": "rimraf build && truffle compile"
   },
   "devDependencies": {
+    "@types/bignumber.js": "^4.0.3",
```
tibetsprague: we dont need this anymore, this is for arc-js

package.json
orenyodfat: Fixed at d4aa48d

package.json
orenyodfat: Fixed at d4aa48d

package.json
orenyodfat: Fixed at d4aa48d

orenyodfat:
Thanks @dkent600 @tibetsprague @leviadam .
Lets approve it.

dkent600:
@orenyodfat You changed the node.js version to 9.3.0.  Just for the record, what was the reason for that?

Does this mean the team now needs to update their node installation to do contract compilation/migrations and/or tests?

orenyodfat:
> @orenyodfat You changed the node.js version to 9.3.0. Just for the record, what was the reason for that?

Yarn was not pleased with node 7.10.0 version : 

> yarn run v1.3.2
warning You are using Node "7.10.0" which is not supported and may encounter bugs or unexpected behavior. Yarn supports the following semver range: "^4.8.0 || ^5.7.0 || ^6.2.2 || >=8.0.0" 

> Does this mean the team now needs to update their node installation to do contract compilation/migrations and/or tests?

This effect only travis. 
If you would like your system to be align with travis ..you can update your node.
Though it is not forced.
In order to force the node version which the npm runs on we can add engine to the package.json.
https://docs.npmjs.com/files/package.json#engines
what do you say ?

Which node version you are using now?

dkent600:
@orenyodfat 

> Which node version you are using now?

I'm using v8.1.2.  Can't speak for everyone else.

> In order to force the node version which the npm runs on we can add engine to the package.json.

Looks like it doesn't force, rather is advisory.  I like that idea.

orenyodfat:
I approve this pr which was started by @dkent600

PR #223, opened 2018-01-05
Update migration script

orenyodfat:
As deployment should be done on higher layers of DAOStack .
DAOStack organization can be also instance on the library level.


leviadam:
I think it is a good idea that the tests will not rely on the migration.
But the migration itself is also an important test.
We have to make sure though that it is clear this is not the DAOstack, but a test and a reference for migration.

orenyodfat:
Thanks @leviadam  for your  comment .
I updated the branch.

Issue #226, opened 2018-01-05
setScheme is reverting when scheme is obtained from .deployed()


dkent600:
setScheme is reverting when the supplied scheme has been obtained using .deployed(), rather than .new.

While there is another issue, #225, where setScheme is reverting when the input token is not equal to the schemes' native token, this issue is different.

#225, however, gives an example of the code that is pertinent to this case:

> If I try to set a scheme creating like this, it works:
> 
> `schemeToAdd = await SchemeRegistrar.new(standardTokenMock.address, 10, accounts[0])`
> 
> (This is how it is done in the arc automated tests.)
> 
> However, if I try to set a scheme creating like this, it reverts:
> 
> `schemeToAdd = SchemeRegistrar.at((await SchemeRegistrar.deployed()).address))`
>
> (This is how it is done in arc-js.)

It is possible to overcome the issue described in #225 by passing the correct tokens into setScheme.  

Even then, setScheme still reverts when using .deployed() to supply the desired schemes.

It reverts on this line:

`UniversalScheme(scheme).registerOrganization(_avatar);`

as can be ascertained by deleting the line and seeing the test succeed.

The weird thing is, one can remove all the code in `UniversalScheme(scheme).registerOrganization` and setSchemes **still** reverts.

So to my eyes there is something wrong with this part: `UniversalScheme(scheme)`.

That's all I got.

orenyodfat:
@dkent600 with which params you are deploying the scheme ?
Could you please write a simple test code which address the issue you are referring to ?

dkent600:
Can't say for sure, but it is possible this is also happening in `SchemeRegistrar.proposeScheme`.

orenyodfat:
Could you please write a small test to clarify your claims.

dkent600:
@orenyodfat 

Extract the attached zip into the test folder.

Then run `truffle test .\test\genesisscheme.js`

[test.zip](https://github.com/daostack/daostack/files/1608872/test.zip)

orenyodfat:
Thanks @dkent600 .
Please remember that you need first to transfer funds to the org avatar for the scheme registration with the same token that the scheme is using.
You are doing that ..but no to the scheme's token.
Please replace :  
> await org.token.transfer(avatarAddress, totalFees, { from: founders[0].address }); 

with : 

> var schemeRegistrarAddress = await SchemeRegistrar.at((await SchemeRegistrar.deployed()).address);
[...]

PR #227, opened 2018-01-05
Include contracts in the repo for use by arc-js

tibetsprague:
I've been playing with way too many options for including accessing contracts in arc-js and especially ones deployed to Kovan, but decided we could start with this incredibly simple option. This just includes the contracts in arc, with kovan deployment addresses, and exports a function to load the contracts. Super basic, but works for now and we just need something working that we agree to move forward with while we think through better options


orenyodfat:
@tibetsprague 
-could you please share the difficulties to use our npm package deploy the contracts in kovan ?
- Could you please share the lib code which trying to do that ? 
- are you sure you need to increase the gas limit ? I succeed migrate to kovan without increasing the gas.

tibetsprague:
@orenyodfat I am not having any difficulties right now, I was able to successfully deploy to Kovan. at one point I did need to increase the gas limit. its going to need to constantly change. in arc-js my latest PR allows you to change it through various config options, we could allow that in arc too.

index.js
```diff
@@ -0,0 +1,3 @@
+export function requireContract(contractName) {
```
dkent600: JavaScript in arc?  Hmm...  ;-)  Anyway, what is this needed for?  arc-js has it's own requireContract.  BTW, we should probably sync up on my work on https://github.com/daostack/arc-js/issues/17

migrations/2_deploy_DAOstack.js
```diff
@@ -55,7 +55,7 @@ module.exports = async function(deployer) {
     // apparently we must wrap the first deploy call in a then to avoid
     // what seem to be race conditions during deployment
     // await deployer.deploy(GenesisScheme)
-    deployer.deploy(GenesisScheme, {gas: 6500000}).then(async function(){
+    deployer.deploy(GenesisScheme, {gas: 6900000}).then(async function(){
```
dkent600: As of today, migrations/2_deploy_DAOstack.js is to be moved to arc-js.  See https://github.com/daostack/daostack/pull/223.

index.js
dkent600: Just to reiterate what you already know:  The arc json contracts files are not supposed to contain migration addresses.  I think you view this as potentially just a work-around.  Just a heads up that a couple things, described in other comments here, don't sync with what I'm doing in arc-js (https://github.com/daostack/arc-js/issues/17), which is currently blocked by #226.  Hopefully that will be resolved very soon.

PR #231, opened 2018-01-09
Add universal controller + Remove universal scheme registration fee

orenyodfat:
-Add universal controller + tests.
-Add ControllerInteface which both universal and none universal controllers should be align with.
-Update genesisscheme to support forging orgs for both universal and none universal controllers.
Note : This update include functions signatures (apis) changes of genesisscheme and controller.
-Fix some linter security warnings while trying to assign values to function parameters.
-Remove universal scheme registration fee.
-decrease gas limit to 6015000.


contracts/controller/Controller.sol
```diff
@@ -397,11 +396,11 @@ contract Controller {
     * @dev decrease approval for the spender address to spend a specified amount of tokens
     *      on behalf of msg.sender.
     * @param _externalToken the address of the Token Contract
-    * @param _spender address
+    * @param _spender addresså
```
leviadam: Typo?

orenyodfat: fixed at 9aaf222

Issue #241, opened 2018-01-12
Problems running tests on windows .


dkent600:
I am having a couple of problems running the tests in Arc.

1) `$npm_package_config_gasLimit` is not resolving to the number that it should.  This prevents migrations from proceeding in arc-js.
2) running `npm run test` produces the error message:  "Could not find ganache-cli", even though it is installed as a package in arc, and I have it also installed globally (though I don't think the scripts should be relying on the latter).

dkent600:
Using this commit:  https://github.com/daostack/daostack/commit/58c04ea096cee43a89721db757ca35e83fb74d76

orenyodfat:
This seems to be a windows issue  : 
For Windows users, environment variables follow a slightly different format, forgoing the $ and instead wrapping the var in %'s like so %npm_package_config_source% .
I will fix that soon.

dkent600:
This could help:  https://www.npmjs.com/package/cross-conf-env

orenyodfat:
yes..
Could you please handle that? as I do not have a window machine to test with currently.

dkent600:
Sure, in the meantime, what about item #2?

orenyodfat:
> Sure, in the meantime, what about item #2?

Seems like another windows issue.

dkent600:
As far as I can tell, the ganache-cli problem is a bug in run-in-testrpc.  I created an issue on their site:  https://github.com/gnosis/run-with-testrpc/issues/6

dkent600:
I really should have created these as separate issues, was lazy, sorry.

I have the fix for the problem with `npm_package_config_gasLimit` and will go ahead and PR it, since that will be sufficient to unblock arc-js.

The problem with ganache-cli only affects being able to run arc tests and does not immediately concern arc-js.

dkent600:
I created #245 for the ganache-cli part of this.

PR #243, opened 2018-01-13
Fix  probem resolving npm_package_config_gasLimit on windows

dkent600:
Fix for item 1 of #241 

This is just grabbing the values from the config entry, not looking at environment variables.

Can we please publish a new NPM package for this?  It is blocking progress on https://github.com/daostack/arc-js/pull/23  (at least, on my machine).  I can work-around it in both arc-js and alchemy, but it's a pain.


package.json
```diff
@@ -81,5 +83,7 @@
     "url": "https://github.com/daostack/daostack/issues"
   },
   "homepage": "https://daostack.io",
-  "dependencies": {}
-}
+  "dependencies": {
+    "cross-conf-env": "^1.1.2"
+  }
+}
```
orenyodfat: no new line at the end of file

PR #256, opened 2018-01-16
Not throwing on negative reputation

leviadam:
This PR changes the logic for decreasing more reputation than available. 
In the past this scenario would revert. Now it will turn the reputation of this agent to zero.
Adapted the tests to the new logic.

In addition I have added call for the event for mint reputation.
Also adjusted the gas limit and the tests accordingly and updated solium version.


contracts/controller/Reputation.sol
```diff
@@ -51,8 +51,12 @@ contract Reputation is Ownable {
             balances[_to] = balances[_to].add(absAmount);
         } else {
             absAmount = uint((-1)*_amount);
-            totalSupply = totalSupply.sub(absAmount);
-            balances[_to] = balances[_to].sub(absAmount);
+            if (absAmount > balances[_to]) {
+                absAmount = balances[_to];
+                _amount = (-1)*int(balances[_to]);
```
orenyodfat: > amount = (-1)*int(balances[_to]);

 do we need this line ?

contracts/controller/Reputation.sol
```diff
@@ -51,8 +51,12 @@ contract Reputation is Ownable {
             balances[_to] = balances[_to].add(absAmount);
```
orenyodfat: @leviadam I guess that the use of safe math to protect against an overflow.right ?
So why not just test that here?.if it get to uint MAX_VALUE just set it to that..avoiding the call to safeMath.

leviadam: I think using safeMath is cleaner.

orenyodfat:
If there is no constraint on the rep so max rep is max uint .
One should calculate the exact rep to add in order to get to that ...
If there is a constraint so it will be revert on the constraint .

leviadam:
@orenyodfat  I don't think we need such a max rep.

docs/README.md
```diff
@@ -1,7 +1,54 @@
-# DAOStack Documentation
+# Resources
```
orenyodfat: How this file is related ?

docs/concepts.md
```diff
@@ -0,0 +1,42 @@
+## An Overview of Arc
```
orenyodfat: How this file is related ?

docs/README.md
leviadam: I have no idea. I thought I rebased.

leviadam:
Strange, I was sure I rebased. Checking.

leviadam:
OK. Closing this and will PR from a new branch to save all the mess.

PR #262, opened 2018-01-18
Use our new package. Fix before-all-hook problem

leviadam:


leviadam:
[Solve “before all hook…” travis error](https://trello.com/c/o0fw8MBd/8-solve-before-all-hook-travis-error)

PR #265, opened 2018-01-19
Refactor reputation contract

leviadam:
1. On negative decreasing more reputation than exists to a user, balnce will be zero instead of throwing.
2. Adjusted the tests and the gas.
3. Updated solium.


leviadam:
[Refactor reputation contract](https://trello.com/c/BR4BwzGG/26-refactor-reputation-contract)

PR #266, opened 2018-01-21
Delete root `README.md`

dev-matan-tsuberi:
Since there's duplication of info in `README.md` and `docs/README.md`, deleted the root one in favor of `docs/README.md`.

**Github should automatically show `docs/README.md` on the main page once merged to master.*


orenyodfat:
@dev-matan-tsuberi I guess it is standard to set the repo readme at the root dir.
docs/README.md should include information about the docs dir. 

If you feel that root readme need to be updated , please open a pr against that so it could be discussed .

dev-matan-tsuberi:
@orenyodfat I think all docs should live under `docs`. 
Making `README.md` an exception has no real benefit since github will serve `docs/README.md` automatically. And if you do so, it only makes links non-relative and will not be served with the rest of the docs if we use serve them using github pages.

orenyodfat:
> And if you do so, it only makes links non-relative and will not be served with the rest of the docs if we use serve them using github pages. 

Can you elaborate more on that ?

And this PR introduce a change in the project readme file. (docs/readme is different from root/readme) 
Please pr your changes.

leviadam:
Hi @dev-matan-tsuberi, the `readme.md` in the master is a standard that many other app use, e.g. the npm website. So I think we have to keep it.

PR #274, opened 2018-01-22
bumped node version to at least 9.3.0

dev-matan-tsuberi:
to support `es6` syntax.

### DIFF
--- a/package.json
+++ b/package.json
@@ -74,7 +74,7 @@
     "arc"
   ],
   "engines": {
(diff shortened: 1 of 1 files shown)


package.json
```diff
@@ -74,7 +74,7 @@
     "arc"
   ],
   "engines": {
-    "node": ">=8.0.0"
+    "node": ">=9.4.0"
```
leviadam: Travis does not support this version.
Please take 9.3.

PR #276, opened 2018-01-24
Add governance contract

orenyodfat:
This PR creates a new "Governance" contract, "GovernanceFormulasInterface" interface and tests.

The governance contract is an advanced voting machine which will be use by daostack's MVP app.
A proposals which maintain by this contracts might have 4 states: 

1.  Not boosted 
-  Proposal is votable.
-  Proposal is stakable.
-  If a proposal gets to absolute majority of the votes it is executed.
-  If a proposal passes the not-boosted time limit (without an absolute votes) it is executed as
 abstained.
 -  If a proposal passes a certain score threshold it shifts to boosted state.
-  Not redeemable 

2. Boosted 
- Proposal is votable.
- Proposal is not stakable.
-  If a proposal passes the boosted time limit it is executed according to a relative majority.
-  Not redeemable 

3. Executed
 - Proposal is not votable
 - Proposal is not stakable.
 - Proposal is redeemable.

4. Doesn't exist

The contracts include a default formula for proposal score, organization threshold and redeemAmount calculation. There is an option to override these formulas using a GovernanceFormulasInterface contract which can be passed as a param to the governance contract.

For a detail specification see doc  (TODO - add link to doc here).

The PR includes some basic tests for this contract though there is always place for additional more complex scenarios tests.


leviadam:
[New voting machine](https://trello.com/c/ehjv6Mk8/6-new-voting-machine)

.soliumrc.json
```diff
@@ -4,6 +4,9 @@
     "security"
   ],
   "rules": {
+    "security/no-block-members": [
+      "off"
```
dev-matan-tsuberi: Why are we turning off this lint?

leviadam: Because we must use the now command.
The ideal thing would be to turn it off just for a specific file or line, but we couldn't find if it is even possible.

orenyodfat: I found a way..will update soon.

orenyodfat: Fixed at a2c5a54

orenyodfat:
Thanks @leviadam & @dev-matan-tsuberi for your comments.
a2c5a54  
 - Disable solium warnings for  specific code lines and remove the need to set its off for all the project.
 - Fix all project solium warnings...which related to functions ordering and use of low level solidity calls.
11d5a69  - Use enum for proposal state.

contracts/VotingMachines/Governance.sol
```diff
@@ -0,0 +1,509 @@
...
+
+
+contract Governance is IntVoteInterface,UniversalScheme,GovernanceFormulasInterface {
+    using SafeMath for uint;
+
+    enum ProposalState { Closed, Executed, NoneBoosted, Boosted }
+
+    struct Parameters {
+        Reputation reputationSystem; // the reputation system that is being used
+        uint noneBoostedVoteRequierePercentage; // the absolute vote percentages bar.
```
dkent600: Misspelled, should be nonBoostedVoteRequiredPercentage?

contracts/VotingMachines/Governance.sol
```diff
@@ -0,0 +1,509 @@
...
+        uint noneBoostedVotePeriodLimit; //the time limit for a proposal to be in an absolute voting mode.
```
dkent600: Misspelled, should be nonBoostedVotePeriodLimit?

contracts/VotingMachines/Governance.sol
```diff
@@ -0,0 +1,509 @@
...
+        uint boostedVotePeriodLimit; //the time limit for a proposal to be in an relative voting mode.
+        uint scoreThreshold;
+        GovernanceFormulasInterface governanceFormulasInterface;
+        uint minimumStaketingFee;
```
dkent600: Misspelled, should be minimumStakingFee?

contracts/VotingMachines/Governance.sol
```diff
@@ -0,0 +1,509 @@
...
+import "./IntVoteInterface.sol";
+import "../universalSchemes/UniversalScheme.sol";
+import "./GovernanceFormulasInterface.sol";
+
+
+
+contract Governance is IntVoteInterface,UniversalScheme,GovernanceFormulasInterface {
+    using SafeMath for uint;
+
+    enum ProposalState { Closed, Executed, NoneBoosted, Boosted }
```
dkent600: Misspelled, should be NotBoosted?

contracts/VotingMachines/Governance.sol
```diff
@@ -0,0 +1,509 @@
...
+        require((proposals[_proposalId].state == ProposalState.NoneBoosted) || (proposals[_proposalId].state == ProposalState.Boosted));
+        _;
+    }
+
+    /**
+     * @dev hash the parameters, save them if necessary, and return the hash value
+    */
+    function setParameters(
+        Reputation _reputationSystem, // the reputation system that is being used
+        uint _noneBoostedVoteRequierePercentage, // the absolute vote percentages bar.
```
dkent600: Misspelled, should be _nonBoostedVoteRequiredPercentage?

contracts/VotingMachines/Governance.sol
```diff
@@ -0,0 +1,509 @@
...
+        uint _noneBoostedVotePeriodLimit, //the time limit for a proposal to be in an absolute voting mode.
```
dkent600: Misspelled, should be _nonBoostedVotePeriodLimit?

contracts/VotingMachines/Governance.sol
```diff
@@ -0,0 +1,509 @@
...
+        uint _boostedVotePeriodLimit, //the time limit for a proposal to be in an relative voting mode.
+        uint _scoreThreshold,
+        GovernanceFormulasInterface _governanceFormulasInterface,
+        uint _minimumStaketingFee)
```
dkent600: Misspelled, should be _minimumStakingFee?

contracts/VotingMachines/Governance.sol
```diff
@@ -0,0 +1,509 @@
...
+        uint boostedPhaseTime; //the time the proposal shift to relative mode.
+        ProposalState state;
+        uint winningVote; //the winning vote.
+        mapping(uint=>uint) votes;
+        mapping(address=>Voter) voters;
+        mapping(uint=>uint) stakes;
+        mapping(address=>Staker) stakers;
+    }
+
+    event NewProposal(bytes32 indexed _proposalId, uint _numOfChoices, address _proposer, bytes32 _paramsHash);
```
dkent600: These events have a different naming convention than `AbsoluteVote`.  I think they should be consistent.  AbsoluteVote events (and I think those in other contracts as well) have the word "Log" prepended, like this:

```
event LogNewProposal(bytes32 indexed _proposalId, uint _numOfChoices, address _proposer, bytes32 _paramsHash);
    event LogCancelProposal(bytes32 indexed _proposalId);
    event LogExecuteProposal(bytes32 indexed _proposalId, uint _decision);
    event LogVoteProposal(bytes32 indexed _proposalId, address indexed _voter, uint _vote, uint _reputation, bool _isOwnerVote);
    event LogCancelVoting(bytes32 indexed _proposalId, address indexed _voter);
```

contracts/VotingMachines/Governance.sol
```diff
@@ -0,0 +1,509 @@
+pragma solidity ^0.4.18;
+
+import "../controller/Reputation.sol";
+import "./IntVoteInterface.sol";
+import "../universalSchemes/UniversalScheme.sol";
+import "./GovernanceFormulasInterface.sol";
+
+
+
+contract Governance is IntVoteInterface,UniversalScheme,GovernanceFormulasInterface {
```
dkent600: Should this have comments describing basically what Governance does?

contracts/VotingMachines/Governance.sol
```diff
@@ -0,0 +1,509 @@
...
+        return proposalId;
+    }
+
+  /**
+   * @dev Cancel a proposal, only the owner can call this function and only if allowOwner flag is true.
+   * @param _proposalId the proposal ID
+   */
+    function cancelProposal(bytes32 _proposalId) public onlyProposalOwner(_proposalId) votable(_proposalId) returns(bool) {
+        //This is not allowed.
+        return false;
```
dkent600: Should this revert instead of returning false, or perhaps add an `allowed(false)` modifier?

contracts/VotingMachines/Governance.sol
```diff
@@ -0,0 +1,509 @@
...
+
+  /**
+   * @dev voting function with owner functionality (can vote on behalf of someone else)
+   * @param _proposalId id of the proposal
+   * @return bool true - the proposal has been executed
+   *              false - otherwise.
+   */
+    function ownerVote(bytes32 _proposalId, uint , address ) public onlyProposalOwner(_proposalId) votable(_proposalId) returns(bool) {
+      //This is not allowed.
+        return false;
```
dkent600: Should this revert instead of returning false, or perhaps add an `allowed(false)` modifier?

contracts/VotingMachines/Governance.sol
```diff
@@ -0,0 +1,509 @@
...
+        return internalVote(_proposalId,msg.sender,_vote,_rep);
+    }
+
+  /**
+   * @dev Cancel the vote of the msg.sender: subtract the reputation amount from the votes
+   * and delete the voter from the proposal struct
+   * @param _proposalId id of the proposal
+   */
+    function cancelVote(bytes32 _proposalId) public votable(_proposalId) {
+       //this is not allowed
```
dkent600: Should this revert instead of returning false, or perhaps add an `allowed(false)` modifier?

contracts/VotingMachines/Governance.sol
```diff
@@ -0,0 +1,509 @@
...
+        if (params.governanceFormulasInterface == GovernanceFormulasInterface(0)) {
+            Proposal storage proposal = proposals[_proposalId];
+            return (proposal.totalStakes * (proposal.totalVotes**2))/(params.reputationSystem.totalSupply()**2);
+        } else {
+            return (params.governanceFormulasInterface).score(_proposalId);
+        }
+    }
+
+    /**
+     * @dev threshold return the organization none boosted threshold
```
dkent600: misspelling: threshold return the organization nonBoosted threshold

contracts/VotingMachines/Governance.sol
```diff
@@ -0,0 +1,509 @@
...
+    /**
+      * @dev isVotable check if the proposal is votable
+      * @param _proposalId the ID of the proposal
+      * @return bool true or false
+    */
+    function isVotable(bytes32 _proposalId) public constant returns(bool) {
+        return ((proposals[_proposalId].state == ProposalState.NoneBoosted) || (proposals[_proposalId].state == ProposalState.Boosted));
+    }
+
+    function proposalStatus(bytes32 _proposalId) public constant returns(uint, uint) {
```
dkent600: Several of the methods below are lacking comments.

contracts/VotingMachines/GovernanceFormulasInterface.sol
```diff
@@ -0,0 +1,14 @@
+pragma solidity ^0.4.18;
+
+
+interface GovernanceFormulasInterface {
+
+    function isBoost(bytes32 _proposalId) public view returns (bool);
```
dev-matan-tsuberi: isBoost -> isBoosted ?

contracts/VotingMachines/Governance.sol
orenyodfat: cancelProposal is not allowed for this votingMachine ( yet).
It is there just because of the voting interface.
Not sure it should revert. 
One should not call that and if it does it will just do nothing.

contracts/VotingMachines/Governance.sol
orenyodfat: ownerVote is not allowed for this votingMachine ( yet).
It is there just because of the voting interface.
Not sure it should revert.
One should not call that and if it does it will just do nothing.

contracts/VotingMachines/Governance.sol
orenyodfat: cancelVote is not allowed for this votingMachine ( yet).
It is there just because of the voting interface.
Not sure it should revert.
One should not call that and if it does it will just do nothing.

contracts/VotingMachines/GovernanceFormulasInterface.sol
orenyodfat: I guess it should be isBoost as it check if a certain proposal should shift to a boost state.

contracts/VotingMachines/Governance.sol
orenyodfat: Will change AbsoluteVote  events as we do not need the "Log" .It is obvious that it is an event.

contracts/VotingMachines/Governance.sol
orenyodfat: Update at a6ba131 .
Though still need more detail doc/description.

contracts/VotingMachines/Governance.sol
orenyodfat: Update at a6ba131

contracts/VotingMachines/Governance.sol
orenyodfat: Update at a6ba131

contracts/VotingMachines/Governance.sol
orenyodfat: Update at a6ba131

contracts/VotingMachines/Governance.sol
orenyodfat: Update at a6ba131

contracts/VotingMachines/Governance.sol
orenyodfat: Update at a6ba131

contracts/VotingMachines/Governance.sol
orenyodfat: Update at a6ba131

contracts/VotingMachines/Governance.sol
orenyodfat: Update at a6ba131

contracts/VotingMachines/Governance.sol
dkent600: Reverting (or maybe better, using a modifier like allowed(false)) would instantly and unequivocally apprise the user that they have done something wrong.

Returning false, unless they are checking the return value, would be silent, thus potentially delaying the time until which the user would realize they've done something wrong.

contracts/VotingMachines/GovernanceFormulasInterface.sol
dkent600: Then "shouldBoost" would be better.

contracts/VotingMachines/Governance.sol
dkent600: I think we should make this change everywhere.   So I created a separate issue for it:  

https://github.com/daostack/Arc/issues/282

orenyodfat: ok. not much left

contracts/VotingMachines/Governance.sol
orenyodfat: I guess it is a good practice to check return value for each call.
revert for this case seems too heavy tool.

PR #286, opened 2018-01-30
Update the different networksId

### LINKED ISSUE #212: prepare truffle.js for mainnet
prepare truffle.js for mainnet, see guidance:  http://truffleframework.com/tutorials/deploying-to-the-live-network

leviadam:
This PR updates the network ID for all common networks.

Resolve #212 .

### DIFF
--- a/truffle.js
+++ b/truffle.js
@@ -11,7 +11,7 @@ module.exports = {
       host: "localhost",
       port: 8546,
       gas: 4543760
-    }
+    },
     ropsten: {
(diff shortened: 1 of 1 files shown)


orenyodfat:
@leviadam is that been tested on rinkeby ,ropsten and kovan ?

leviadam:
I copied it. I don't think it requires tests.
It is just a shortcut for truffle from name to network id, and those I double checked.

PR #306, opened 2018-02-02
Update Repo Url

Alonski:
Should be arc.git and not Arc.git


dev-matan-tsuberi:
Please squah before merge.

PR #307, opened 2018-02-04
Updated Packages and Fixed Failing Tests

Alonski:
Updated NPM Packages
In Truffle v4.0.6 there was a breaking change that changes how contracts can be created with .new()
https://github.com/trufflesuite/truffle-contract/issues/82
I updated all tests that were causing issues to be compliant with this.

In addition I tried to update bignumber.js to v6.0.0. This also introduced some breaking changes:
https://github.com/MikeMcl/bignumber.js/blob/master/CHANGELOG.md

However web3 hasn't been updated for this. 
I can create a PR for web3 to fix this.


orenyodfat:
@Alonski thanks.
Truffle v4.0.6  also update solc to 4.0.19 .
All of our contracts are set to use 4.0.18 .
Could you please try to update all our *.sol files to use solc 4.0.19 ?

Alonski:
@orenyodfat No problem
I knew you would ask about updating everything to 4.0.19 :)
I tried and found that there is a problem with the zeppelin contracts. They won't compile for some reason.

There was also a linting error that I am fixing.
Updating ESLint to v4.17 was also causing lint to fail sometimes. I am reverting back to v4.16

Alonski:
Zeppelin changed folder structure without declaring a breaking change :)
Fixing

PR #315, opened 2018-02-05
Controller global constraint optimization.

orenyodfat:
Controllers will maintain globalConstraintsPre and globalConstraintsPost arrays (instead of globalConstraints). This will make onlySubjectToConstraint to be more efficient.


orenyodfat:
@dev-matan-tsuberi 
 > Shouldn't we make two interfaces GlobalConstraintPreInterface,GlobalConstraintPostInterface? 

Why ?

dev-matan-tsuberi:
@orenyodfat 
I just thought there should two types of global constraints, pre & post. you can register them independently. The post global constraints work as I said (they maintain the last state they saw). and the pre global constraints are just non-stateful predicates just on the current state (for the efficiency reasons you mentioned).

Thought that this was what you were implementing.

contracts/globalConstraints/GlobalConstraintInterface.sol
```diff
@@ -2,6 +2,14 @@ pragma solidity ^0.4.19;
 
 
 contract GlobalConstraintInterface {
+
+    enum CallPhase { Never, Pre, Post,PreAndPost }
```
dev-matan-tsuberi: Why we need `Never`?

dev-matan-tsuberi:
https://github.com/daostack/arc/issues/316

contracts/globalConstraints/GlobalConstraintInterface.sol
orenyodfat: Well, I still do not really knows :). 
Though I can think on a scenario which one might want to use a dynamic global constraint contract ...which is controlled by some entity which can turn this gc contract to none valid without the need to remove it from the controller....

Though it might be useless and should be removed.

dev-matan-tsuberi: What's wrong with adding/removing on demand? Using the `Never` approach requires special logic for permissions etc...

orenyodfat: Removed at 64c3a9a

PR #332, opened 2018-02-07
Genesisprotocol : Make it YES/NO voting only

orenyodfat:
- Currently there is only  YES/NO voting choices. (NUM_OF_CHOICES is 2).
- Abstain (0) - cannot be voted.
- properly delete proposal on quorum vote.
- VotesStatus interface function changed to VoteStatus.

### DIFF
--- a/contracts/VotingMachines/AbsoluteVote.sol
+++ b/contracts/VotingMachines/AbsoluteVote.sol
@@ -206,15 +206,13 @@ contract AbsoluteVote is IntVoteInterface {
     }
 
     /**
-     * @dev votesStatus returns the number of yes, no, and abstain and if the proposal is ended of a given proposal id
+     * @dev voteStatus returns the reputation voted for a proposal for a specific voting choice.
      * @param _proposalId the ID of the proposal
-     * @return votes array of votes for each choice
+     * @param _choice the index in the
+     * @return voted reputation for the given choice
      */
-    function votesStatus(bytes32 _proposalId) public view returns(uint[11] votes) {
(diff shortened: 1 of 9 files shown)


orenyodfat:
@dkent600 
re 

> Why have MAX_NUM_OF_CHOICES if it can't ever be other-than 2?
Should the _numOfChoices argument in propose be removed?

It is NUM_OF_CHOICES  not MAX_NUM_OF_CHOICES 
propose is an interface function for IntVoteInterface . It keeping the same interface.

orenyodfat:
> The caller of voteStatus needs to be able to know how many items are in the returned array.
If voteStatus is going to always return an array with MAX_NUM_OF_CHOICES + 1 items (which make no sense to me in any case), then the caller needs to know the value of MAX_NUM_OF_CHOICES.
But MAX_NUM_OF_CHOICES is not public.
Perhaps this is moot now, but if you're going to retain the code that makes it seem like MAX_NUM_OF_CHOICES could someday change, that makes it seem like there is truly a choice in proposal.numOfChoices (though at the moment I don't think truly is) then I think we need to give the user some way to know MAX_NUM_OF_CHOICES at runtime.
(But I think a better change would to make voteStatus always return an array of size proposal.numOfChoices + 1.)

Please take a look on carefully on the pr .
voteStatus does not return array.

dkent600:
Sorry, indeed I hadn't looked at all the changes.

dkent600:
I do have another question:  Should NO and YES be public?

dkent600:
Do we still need `getNumberOfChoices`?

PR #348, opened 2018-02-13
avatar : revert if delegatecall return false

orenyodfat:
Due to the fact that delegatecall method never throw an exception, but will return false if the call encounters an exception.

### DIFF
--- a/contracts/controller/Avatar.sol
+++ b/contracts/controller/Avatar.sol
@@ -58,10 +58,12 @@ contract Avatar is Ownable {
     {
         GenericAction(_action, _params);
         // solium-disable-next-line security/no-low-level-calls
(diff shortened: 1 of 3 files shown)


contracts/controller/Avatar.sol
```diff
@@ -58,10 +58,12 @@ contract Avatar is Ownable {
     {
         GenericAction(_action, _params);
         // solium-disable-next-line security/no-low-level-calls
-        return _action.delegatecall(bytes4(keccak256("action(bytes32[])")),
+        bool ret = _action.delegatecall(bytes4(keccak256("action(bytes32[])")),
         uint256(32),// length of length of the array
         uint256(_params.length), // length of the array
         _params);                 // array itself);
+        require(ret);
+        return ret;
```
leviadam: `return true`?

PR #349, opened 2018-02-13
Set a fixed version of solidity without the caret

### LINKED ISSUE #343: set a fixed version of solidity without the caret e.g 0.4.19.
Specifying a pragma version with the caret symbol (^) up front tells the compiler to use any version of solidity bigger than the version mentioned  .
This is not a good practice since there could be major changes between versions that would make the code unstable.

orenyodfat:
resolve #343


dev-matan-tsuberi:
See https://github.com/daostack/arc/issues/343 discussion.

PR #357, opened 2018-02-14
Moved `zeppelin-solidity` to deps, moved `cross-conf-env` to devDeps

dev-matan-tsuberi:
Since we want users to import `.sol` contracts and compile them from source, we need to have `zeppelin-solidity` as a dep and not devDep, otherwise users would not be able to compile.

Also `cross-conf-env` should be a devDep.

**Please republish npm package after merging this PR.**

### DIFF
--- a/package.json
+++ b/package.json
@@ -37,6 +37,7 @@
     "babel-preset-es2015": "^6.24.1",
     "babel-register": "^6.26.0",
     "bignumber.js": "^5.0.0",
+    "cross-conf-env": "^1.1.2",
     "default-options": "^1.0.0",
     "eslint": "^4.17.0",
     "eslint-config-defaults": "^9.0.0",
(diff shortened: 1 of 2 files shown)


package.json
```diff
@@ -62,8 +63,7 @@
     "web3-typescript-typings": "^0.9.8",
     "web3-utils": "^1.0.0-beta.29",
     "when": "^3.7.8",
-    "yargs": "^11.0.0",
-    "zeppelin-solidity": "^1.6.0"
+    "yargs": "^11.0.0"
```
orenyodfat: do we need yargs ?

dev-matan-tsuberi: not anymore actually, since https://github.com/daostack/arc/pull/347

dev-matan-tsuberi: updated: https://github.com/daostack/arc/pull/357/commits/3acbe39c04b10f6f5699e96068865757c15aaea7

dev-matan-tsuberi:
@orenyodfat Can you please republish the npm package?

PR #369, opened 2018-02-19
GenesisProtocol enhancement 

orenyodfat:
- Add validity checks for orgs params. 
- Require stake amount > 0.
- Tests.
- Refactor execute function.


dev-matan-tsuberi:
Consider specifying those `require`ments in the docs strings, so that people are aware.

PR #377, opened 2018-02-21
Add _param to ProposalExecution  events

### LINKED ISSUE #372: `ChangeUpgradeSchemeProposal` `newUpgradeScheme` parameter should be preceded with an underscore
To be consistent with every other event parameter in Arc, the `ChangeUpgradeSchemeProposal` `newUpgradeScheme` parameter should be preceded with an underscore.

### LINKED ISSUE #376: UpgradeScheme and VestingScheme don't fire ProposalExecuted
Both UpgradeScheme and VestingScheme declare ProposalExecuted but don't fire it.

Further:  VestingScheme declares it without _proposalId as a parameter.  _proposalId is a parameter on that event everywhere else and we should be consistent.

orenyodfat:
GenesisProtocol : extend ExecuteProposal event.

resolve #376 
resolve #372

### DIFF
--- a/contracts/VotingMachines/GenesisProtocol.sol
+++ b/contracts/VotingMachines/GenesisProtocol.sol
@@ -13,6 +13,8 @@ contract GenesisProtocol is IntVoteInterface,UniversalScheme,GenesisProtocolForm
     using SafeMath for uint;
 
     enum ProposalState { Closed, Executed, PreBoosted,Boosted,QuietEndingPeriod }
+    enum ExecutionState { None, PreBoostedTimeOut, PreBoostedBarCrossed, BoostedTimeOut,BoostedBarCrossed }
+
     //Organization's parameters
     struct Parameters {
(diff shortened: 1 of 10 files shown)


contracts/VotingMachines/GenesisProtocol.sol
```diff
@@ -66,7 +68,7 @@ contract GenesisProtocol is IntVoteInterface,UniversalScheme,GenesisProtocolForm
     }
 
     event NewProposal(bytes32 indexed _proposalId, uint _numOfChoices, address _proposer, bytes32 _paramsHash);
-    event ExecuteProposal(bytes32 indexed _proposalId, uint _decision);
+    event ExecuteProposal(bytes32 indexed _proposalId, uint _decision,uint _totalReputation,ExecutionState _executionState);
```
dkent600: Should AbsoluteVote have a corresponding change, at least with regard to returning reputation?  And is it possible to keep this event signature consistent across voting machines?

dkent600:
Can we have a new Arc package with this?

PR #389, opened 2018-03-02
extend ERC827 instead of StandardToken

dev-matan-tsuberi:
Increase GAS limit to 6600000
Add more gas to DAOCreatore.forgeOrg
DAOToken is not destructible any more.


orenyodfat:
Increase GAS limit to 6600000
Add more gas to DAOCreatore.forgeOrg
DAOToken is not destructible any more.

Alonski:
Just wondering. Why was destructable removed?

dev-matan-tsuberi:
[Token transfer and call scheme](https://trello.com/c/TMqBadLF/3-token-transfer-and-call-scheme)

dev-matan-tsuberi:
@Alonski 
Because it's unneeded and because it caused the weird error (god knows why).

PR #392, opened 2018-03-05
test for global uniqueness of proposalId's.

dev-matan-tsuberi:


test/votingMachine.js
```diff
@@ -0,0 +1,35 @@
...
+    const token = await StandardToken.new();
+    const rep = await Reputation.new();
+    const absolute = await AbsoluteVote.new();
+    const genesis = await GenesisProtocol.new(token.address);
+    const quorum = await QuorumVote.new(token.address);
+    const avatar = await Avatar.new('name', helpers.NULL_ADDRESS, rep.address);
+    const executable = await ExecutableTest.new();
+
+    const absoluteParams = await absolute.setParameters.call(rep.address,50,true);
+    await absolute.setParameters.sendTransaction(rep.address,50,true);
```
orenyodfat: why sendTransaction and not just setParameters ? 
it will be nice to keep consistency for with all other tests.

test/votingMachine.js
```diff
@@ -0,0 +1,35 @@
...
+
+    const absoluteProposalId = await absolute.propose(5, absoluteParams, avatar.address, executable.address,accounts[0]);
+    const genesisProposalId = await genesis.propose(2, genesisParams, avatar.address, executable.address,accounts[0]);
+    const quorumProposalId = await quorum.propose(5, quoromParams, avatar.address, executable.address,accounts[0]);
+
+    assert(absoluteProposalId !== genesisProposalId, 'AbsoluteVote gives the same proposalId as GenesisProtocol');
+    assert(genesisProposalId !== quorumProposalId, 'GenesisProtocol gives the same proposalId as QuorumVote');
+    assert(quorumProposalId !== absoluteProposalId, 'QuorumVote gives the same proposalId as AbsoluteVote');
+  });
+});
```
orenyodfat: how about adding a test which compare proposalIDs  from two different instances of the same voting machine ?

dev-matan-tsuberi: Yes, this should be done within each voting machine test. It may already be guaranteed by existing tests actually.

orenyodfat: could you please add links to these tests ?

test/votingMachine.js
dev-matan-tsuberi: when you do a `absolute.setParameters(..)` you let web3.js decide whether to `call` or to `sendTransaction`. In this case I think it's better to explicitly write that the first line only get's the result of the function and the second one actually changes the state of the contract.

test/votingMachine.js
dev-matan-tsuberi: I don't know if there are, I said maybe.

dev-matan-tsuberi: For example, I'm pretty sure some tests in `quoromVote.js` would not pass if `proposalIds` were not unique per instance.

test/votingMachine.js
orenyodfat: This is true if you are writing  raw requests via web3.js to the Ethereum network yourself.
Truffle is taking care of that (choosing between 'call' or  'sendTransaction' ) according to what the func is doing (write or read data).
I would ommit  'sendTransaction' for this case.

dev-matan-tsuberi: updated: https://github.com/daostack/arc/pull/392/commits/ae904a0b4448b164af79ca781b211892d8d5cc98

PR #394, opened 2018-03-05
Remove an Unnecessary Import

ritzdorf:
This changes removes the import of Controller from the Avatar. This 
import seems to be unused as Controller is never referenced within
Avatar. Additionally, the circular dependency between Avatar and
Controller can be eliminated by removing this import.

All tests pass.


orenyodfat:
This is already address at #393

ritzdorf:
Sorry.

PR #412, opened 2018-03-14
Absolutevote : add refreshReputation function

orenyodfat:
- Add refreshReputation function can be called by one to refresh the reputation votes for a given proposal and list of voters during voting process. 

- Correct vesting and upgrade schemes permission at test and test migration script.

### DIFF
--- a/contracts/VotingMachines/AbsoluteVote.sol
+++ b/contracts/VotingMachines/AbsoluteVote.sol
@@ -36,6 +36,8 @@ contract AbsoluteVote is IntVoteInterface {
     event ExecuteProposal(bytes32 indexed _proposalId, uint _decision, uint _totalReputation);
     event VoteProposal(bytes32 indexed _proposalId, address indexed _voter, uint _vote, uint _reputation, bool _isOwnerVote);
     event CancelVoting(bytes32 indexed _proposalId, address indexed _voter);
+    event RefreshReputation(bytes32 indexed _proposalId, address indexed _voter,uint _reputation);
+
 
     mapping(bytes32=>Parameters) public parameters;  // A mapping from hashes to parameters
     mapping(bytes32=>Proposal) public proposals; // Mapping from the ID of the proposal to the proposal itself.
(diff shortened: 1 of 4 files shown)


contracts/VotingMachines/AbsoluteVote.sol
```diff
@@ -232,6 +234,37 @@ contract AbsoluteVote is IntVoteInterface {
...
+     */
+    function refreshReputation(bytes32 _proposalId, address[] _voters) public returns(bool) {
+        Proposal storage proposal = proposals[_proposalId];
+        Parameters memory params = parameters[proposal.paramsHash];
+
+        for (uint i = 0; i < _voters.length; i++) {
+            Voter storage voter = proposal.voters[_voters[i]];
+             //check that the voters already votes.
+            if (voter.reputation > 0) {
+                //if the reputation change..so update.
```
leviadam: Fix comment.

contracts/VotingMachines/AbsoluteVote.sol
```diff
@@ -232,6 +234,37 @@ contract AbsoluteVote is IntVoteInterface {
...
+                if (rep > voter.reputation) {
+                    proposal.votes[voter.vote] = proposal.votes[voter.vote].add(rep - voter.reputation);
+                    proposal.totalVotes = (proposal.totalVotes).add(rep - voter.reputation);
+                  } else if (rep < voter.reputation) {
+                    proposal.votes[voter.vote] = proposal.votes[voter.vote].sub(voter.reputation - rep);
+                    proposal.totalVotes = (proposal.totalVotes).sub(voter.reputation - rep);
+                  }
+                if (rep != voter.reputation) {
+                    voter.reputation = rep;
+                    RefreshReputation(_proposalId,_voters[i],rep);
```
leviadam: `return true`?

contracts/VotingMachines/AbsoluteVote.sol
orenyodfat: fixed at 3f8f14f

contracts/VotingMachines/AbsoluteVote.sol
orenyodfat: fixed at 3f8f14f

PR #414, opened 2018-03-19
Security fixes and enhancements 

orenyodfat:
This PR includes the following updates  

- Add ControllerCreator contract as an helper to forgeOrg.
   It is added because DaoCreator cannot be deployed due to limitation tx(deployment size).
   https://github.com/ethereum/EIPs/blob/master/EIPS/eip-170.md 
   gasLimit is set to 5100000
-Controller : Check that passing avatar param is a valid (belong to this controller).
-UController: Check Reputation and DAOToken uniqueness at newOrganization func
-GenesisProtocol : -Additional params validation .
                                 -Fix  rounding calculations.
                                 -Address audit feedback.


test/constants.js
```diff
@@ -1 +1 @@
-export const GENESIS_SCHEME_GAS_LIMIT = 6600000;
+export const GENESIS_SCHEME_GAS_LIMIT = 5100000;
```
dkent600: Should this be 5000000?

orenyodfat:
Will fix that soon

test/constants.js
orenyodfat: 5100000

PR #437, opened 2018-04-12
SimpleICO : require cap!=0 .

orenyodfat:


orenyodfat:
@dev-matan-tsuberi

dev-matan-tsuberi:
Why do we want this?

Issue #440, opened 2018-04-18
Let proposeContributionReward to take an array of beneficiaries


dkent600:
Let `ContributionReward.proposeContributionReward` to take an array of beneficiaries.  This would be used in the current pilot from bringing in the participants who aren't founders.  It would also, in and of itself, be a good test case to start the pilot.

The feature would probably be very useful for DAOs in general who are bringing large groups of people into the DAO all at once.  

This change could be easier and have a lot more leverage than writing a new contract for the one-off case of the pilot.

dkent600:
![](https://github.trello.services/images/mini-trello-icon.png) [Let proposeContributionReward take an array of beneficiaries](https://trello.com/c/yBzSyOTr/139-let-proposecontributionreward-take-an-array-of-beneficiaries)

leviadam:
I don't think this should be under this contract

leviadam:
It is easy to have a beneficiary contract that splits the reward.
And we should modularize and make logic simple when possible.

PR #443, opened 2018-04-21
make ContributionReward handle multiple beneficiaries, scripts work in Windows

dkent600:
Resolves: #440 

- modifies ContributionReward to work with an array of beneficiaries
- adds new `getBeneficiaries` function
- fixes package scripts to work on Windows


orenyodfat:
please see comments in #440

leviadam:
I think this is not a good idea, @orenyodfat already explained.

Issue #452, opened 2018-05-02
Add standard voting machine events to IntVoteInterface


dkent600:
Would be great if `IntVoteInterface` included the standard voting machine events:

NewProposal
CancelProposal
ExecuteProposal
VoteProposal
CancelVoting

With the parameters that are considered standard.

dkent600:
Since GP doesn't do the `Cancel` events, we might leave them out.  Better, I think, would be to add them to GP as stubs even though they are never fired.

Issue #450, opened 2018-04-30
AbsoluteVote.NewProposal should have _avatar parameter


dkent600:
`GenesisProtocol.NewProposal` event has an `_avatar` parameter.

`AbsoluteVote.NewProposal` should too.  We need this to be able to filter the events by avatar.

PR #455, opened 2018-05-09
add events to IntVoteInterface, avatar to those events

dkent600:
Resolves: #450 
Resolves: #452 

I'm uncertain about one aspect of this, which is the mismatch in events between `IntVoteInterface` and `GenesisProtocol` and `AbsoluteVote`:

- GP doesn't declare the `CancelProposal` and `CancelVoting` events
- GP `ExecuteProposal` has an extra param: `ExecutionState _executionState`
- AV `VoteProposal` has an extra param: `bool isOwnerVote` param

Maybe this is all not a problem, but any suggestions are invited about how to better address these differences.

[Edit:]  Decision for above was to add new `AVVoteProposal` and `GPExecuteProposal` events.


contracts/VotingMachines/AbsoluteVote.sol
```diff
@@ -95,11 +95,13 @@ contract AbsoluteVote is IntVoteInterface {
    * @param _proposalId the proposal ID
    */
     function cancelProposal(bytes32 _proposalId) external onlyProposalOwner(_proposalId) votable(_proposalId) returns(bool) {
+
         if (! parameters[proposals[_proposalId].paramsHash].allowOwner) {
             return false;
         }
+        address avatar = proposals[_proposalId].avatar;
```
orenyodfat: how about emit the event before deleting the proposal and save this assignment ?

contracts/VotingMachines/IntVoteInterface.sol
```diff
@@ -9,6 +9,12 @@ interface IntVoteInterface {
     modifier onlyProposalOwner(bytes32 _proposalId) {revert(); _;}
     modifier votable(bytes32 _proposalId) {revert(); _;}
 
+    event NewProposal(bytes32 indexed _proposalId, address indexed _avatar, uint _numOfChoices, address _proposer, bytes32 _paramsHash);
```
orenyodfat: please remove this from the interface.

contracts/VotingMachines/AbsoluteVote.sol
dkent600: Are you sure? The pattern in all the other contracts is to emit *after* deletion.

contracts/VotingMachines/IntVoteInterface.sol
dkent600: Can you please explain why, for the record?

orenyodfat: if it is the intrerface so remove that from the GenesisProtocol and AbsoluteVote

dkent600: Done

dkent600: AbsoluteVote adds a parameter, so I left it in.  Same deal for `GP.ExecuteProposal`.  (See my comments in the PR description.)

orenyodfat: so remove it from the interface

orenyodfat: To many unneeded events.

dkent600: OK, then I would vote to eliminate `AbsoluteVoteProposal` (the existing event in AV) if it is unneeded.  The `IntVoteProposal` version *is* needed.

Similarly with `ExecuteProposal` in GP.

We need to be able to subscribe to these events without knowing what the voting machine is.  You OK with that?

orenyodfat: AbsoluteVoteProposal does not exist in ARC.
you mean VoteProposal ?

dkent600: I describe what I mean by "AbsoluteVoteProposal" in my comment in this thread just above.

dkent600: But yes, I would propose to eliminate the existing `AbsoluteVote.VoteProposal` and `GenesisProtocol.ExecuteProposal` events if you don't like my proposal above to rename and emit them as separate events.

dkent600: @orenyodfat 

This is a long and confusing thread, and it needs to be resolved, so let me summarize where we are at on it:

There is a mismatch in two different events between IntVoteInterface and GenesisProtocol and AbsoluteVote:

* GP.ExecuteProposal has an extra param: ExecutionState _executionState
* AV.VoteProposal has an extra param: bool isOwnerVote param

In order to conform to the IntVoteInterface, AV and GP needs to emit the exact events as they are defined in the interface.  As far as I know, there are two ways to do this:

1) Remove the extra parameters from the extended events in AV and GP
2) Have AV and GP throw both the interface version and extended versions of the events.  We would rename the extended events.

Can you please indicate which is the preferred option, thanks?

contracts/VotingMachines/AbsoluteVote.sol
```diff
@@ -98,11 +94,13 @@ contract AbsoluteVote is IntVoteInterface {
    * @param _proposalId the proposal ID
    */
     function cancelProposal(bytes32 _proposalId) external onlyProposalOwner(_proposalId) votable(_proposalId) returns(bool) {
+
```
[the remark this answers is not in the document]
dkent600: Removed

contracts/VotingMachines/IntVoteInterface.sol
dkent600: Decision is to add new events: AVVoteProposal and GPExecuteProposal.  This is done.

contracts/VotingMachines/AbsoluteVote.sol
```diff
@@ -31,7 +31,7 @@ contract AbsoluteVote is IntVoteInterface {
         bool open; // voting open flag
     }
 
-    event VoteProposal(bytes32 indexed _proposalId, address indexed _avatar, address indexed _voter, uint _vote, uint _reputation, bool _isOwnerVote);
+    event AVVoteProposal(bytes32 indexed _proposalId, address indexed _avatar, address indexed _voter, uint _vote, uint _reputation, bool _isOwnerVote);
```
orenyodfat: the AVVoteProposal should include only the extended params which are not included on VoteProposal interface (with the proposalId as Index).

contracts/VotingMachines/GenesisProtocol.sol
```diff
@@ -80,7 +80,7 @@ contract GenesisProtocol is IntVoteInterface,UniversalScheme {
         mapping(address=>Staker) stakers;
     }
 
-    event ExecuteProposal(bytes32 indexed _proposalId,
+    event GPExecuteProposal(bytes32 indexed _proposalId,
```
orenyodfat: the GPExecuteProposal should include only the extended params which are not included on ExecuteProposal common interface (with the proposalId as Index).

contracts/VotingMachines/AbsoluteVote.sol
```diff
@@ -31,12 +31,8 @@ contract AbsoluteVote is IntVoteInterface {
         bool open; // voting open flag
     }
 
-    event NewProposal(bytes32 indexed _proposalId, uint _numOfChoices, address _proposer, bytes32 _paramsHash);
-    event CancelProposal(bytes32 indexed _proposalId);
-    event ExecuteProposal(bytes32 indexed _proposalId, uint _decision, uint _totalReputation);
-    event VoteProposal(bytes32 indexed _proposalId, address indexed _voter, uint _vote, uint _reputation, bool _isOwnerVote);
-    event CancelVoting(bytes32 indexed _proposalId, address indexed _voter);
-    event RefreshReputation(bytes32 indexed _proposalId, address indexed _voter,uint _reputation);
+    event AVVoteProposal(bytes32 indexed _proposalId, address indexed _avatar, address indexed _voter, uint _vote, uint _reputation, bool _isOwnerVote);
```
orenyodfat: AVVoteProposal should include only the extended param which is not included at the interface event (plus the proposalId as Index)..
There is no need to duplicate .

dkent600: Done

contracts/VotingMachines/GenesisProtocol.sol
dkent600: Done

contracts/VotingMachines/AbsoluteVote.sol
orenyodfat: Why the _avatar  and _voter are needed indexes at the AVVoteProposal ?
One can fetch the proposalId for VoteProposal (by _avatar or _voter) and then look for AVVoteProposal according to _proposalId .
These two events are emitted one after the other.

contracts/VotingMachines/GenesisProtocol.sol
orenyodfat: Does the index  _avatar is really needed here ?
One can fetch the _proposalId via the _avatar from ExecuteProposal .

dkent600: > Does the index _avatar is really needed here?

I left it in to maintain compatibility with all of the other GP events, as well as all of the events in AbsoluteVote and IntVoteInterrface and almost every other scheme event in Arc.

The only inconsistencies are a handful of events in VestingScheme (five) and VoteInOrganizationScheme (one).  I think we should seriously consider adding indexed _avatar to those as well.

This kind of API consistency is very important in supporting consumers of any library or framework, including this one.

contracts/VotingMachines/AbsoluteVote.sol
dkent600: @orenyodfat  This is for API consistency.  Is there a cost to having this indexed _avatar here?

contracts/VotingMachines/AbsoluteVote.sol
dkent600: Removed

contracts/VotingMachines/GenesisProtocol.sol
dkent600: Removed

Issue #451, opened 2018-05-01
supply agreementId with VestingScheme.ProposalExecuted


dkent600:
Can we supply the new agreementId with VestingScheme.ProposalExecuted? Arc.js needs this to be able to connect executed proposals with agreements.

dkent600:
Rather than mess up the standardized parameters of `ProposalExecuted` I decided to add `_proposalId` to `NewVestedAgreement`.  A comprise but workable for Arc.js.

Issue #448, opened 2018-04-25
make VestingScheme.AgreementProposal.ProposalId be indexed 


dkent600:
Can we please make `VestingScheme.AgreementProposal._proposalId` be indexed, so we can filter those events by _proposalId?

PR #456, opened 2018-05-09
proposalId in NewVestedAgreement and indexed proposalId

dkent600:
Resolves: #448 
Resolves: #451

- For 451, rather than mess up the standardized parameters of ProposalExecuted I decided to add a new ProposedVestedAgreement event that contains _proposalId.  This will work for Arc.js.
- indexed _proposalId in AgreementProposal

### DIFF
--- a/contracts/universalSchemes/VestingScheme.sol
+++ b/contracts/universalSchemes/VestingScheme.sol
@@ -239,7 +239,6 @@ contract VestingScheme is UniversalScheme, ExecutableInterface {
             agreements[agreementsCounter] = proposedAgreement;
             agreementsCounter++;
         // Log the new agreement:
-            emit NewVestedAgreement(agreementsCounter-1);
             emit ProposedVestedAgreement(agreementsCounter-1, _proposalId);
         }
         emit ProposalExecuted(_avatar,_proposalId,_param);
--- a/contracts/universalSchemes/VestingScheme.sol
+++ b/contracts/universalSchemes/VestingScheme.sol
@@ -15,8 +15,8 @@ contract VestingScheme is UniversalScheme, ExecutableInterface {
     event ProposalExecuted(address indexed _avatar, bytes32 indexed _proposalId,int _param);
     event ProposalDeleted(address indexed _avatar, bytes32 indexed _proposalId);
     event AgreementProposal(address indexed _avatar, bytes32 indexed _proposalId);
-    // _proposalId will only be set when agreement is created by proposal
-    event NewVestedAgreement(uint indexed _agreementId, bytes32 indexed _proposalId);
+    event NewVestedAgreement(uint indexed _agreementId);
+    event ProposedVestedAgreement(uint indexed _agreementId, bytes32 indexed _proposalId);
     event SignToCancelAgreement(uint indexed _agreementId, address indexed _signer);
     event RevokeSignToCancelAgreement(uint indexed _agreementId, address indexed _signer);
     event AgreementCancel(uint indexed _agreementId);
@@ -180,7 +180,7 @@ contract VestingScheme is UniversalScheme, ExecutableInterface {
         agreementsCounter++;
 
         // Log new agreement and return id:
-        emit NewVestedAgreement(agreementsCounter-1, "");
+        emit NewVestedAgreement(agreementsCounter-1);
         return(agreementsCounter-1);
     }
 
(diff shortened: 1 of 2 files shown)


contracts/universalSchemes/VestingScheme.sol
```diff
@@ -184,7 +185,7 @@ contract VestingScheme is UniversalScheme, ExecutableInterface {
         agreementsCounter++;
 
         // Log new agreement and return id:
-        emit NewVestedAgreement(agreementsCounter-1);
+        emit NewVestedAgreement(agreementsCounter-1, "");
```
orenyodfat: ??? "" ??  not nice.
please revert or add other event without proposalId

dkent600: added an event `ProposedVestedAgreement`

contracts/universalSchemes/VestingScheme.sol
```diff
@@ -239,6 +240,7 @@ contract VestingScheme is UniversalScheme, ExecutableInterface {
             agreementsCounter++;
         // Log the new agreement:
             emit NewVestedAgreement(agreementsCounter-1);
```
orenyodfat: NewVestedAgreement can be remove here..no ?

test/vestingscheme.js
```diff
@@ -115,6 +115,36 @@ contract('VestingScheme', function(accounts) {
...
+                                                                            11,
+                                                                            0,
+                                                                            [],
+                                                                            testSetup.org.avatar.address);
+            //Vote with reputation to trigger execution
+             var proposalId = await helpers.getValueFromLogs(tx, '_proposalId',1);
+             await testSetup.vestingSchemeParams.votingMachine.absoluteVote.vote(proposalId,1,{from:accounts[2]});
+            const found = await new Promise((resolve) => {
+                testSetup.vestingScheme.ProposedVestedAgreement({_proposalId: proposalId}, {fromBlock: 0})
+                    .get((err,events) => {
```
orenyodfat: are you assuming here that there will be only one event (one test on the same test blockchain instance ) from block 0  ? Not sure if it is enough.

contracts/universalSchemes/VestingScheme.sol
dkent600: We can remove it only if you are OK with subtracting from the originally intended functionality of that event, which was to be thrown in both cases where an agreement is created.  Let me know if you're OK with that and I'll delete the line.

test/vestingscheme.js
dkent600: Good idea,  but note that I am filtering the events by the proposalId, which is unique.  So there should only be a single event returned.  Thus I think this is OK for the case you described.

orenyodfat: why you are not getting the event in the normal standard way?

dkent600: By "normal standard way", I assume you mean from the tx log after voting?

It is because `ProposedVestedAgreement` doesn't appear in the tx.logs.  Only `VoteProposal` and `ExecuteProposal` appear there.  Yet `ProposedVestedAgreement` *is* being fired.

contracts/universalSchemes/VestingScheme.sol
orenyodfat: please remove NewVestedAgreement here

test/vestingscheme.js
```diff
@@ -70,7 +70,7 @@ contract('VestingScheme', function(accounts) {
        assert.equal(tx.logs[0].event, "AgreementProposal");
        var avatarAddress = await helpers.getValueFromLogs(tx, '_avatar',1);
        assert.equal(avatarAddress,testSetup.org.avatar.address);
-      });
+ });
```
orenyodfat: this diff is require ?

test/vestingscheme.js
orenyodfat: How about reading the block number of the tx and explicitly set the fromBlock and toBlock with that ?

test/vestingscheme.js
```diff
@@ -219,7 +249,7 @@ contract('VestingScheme', function(accounts) {
                assert.equal(await testSetup.org.token.balanceOf(testSetup.vestingScheme.address),amountPerPeriod*numberOfAgreedPeriods);
               });
 
-              it("createVestedAgreement check agreement id ", async function() {
+              it("createVestedAgreement check agreement", async function() {
```
orenyodfat: this test check the agreement id .why to remove it ?

test/vestingscheme.js
orenyodfat: @dkent600 is that been fixed?

dkent600: yes

test/vestingscheme.js
dkent600: I put it back

test/vestingscheme.js
dkent600: removed

PR #461, opened 2018-05-16
Add dao bounty to genesis protocol

orenyodfat:


leviadam:
Looking at this code I see this protocol is no good.
One cannot just put a cap on the bounty per agent.
A smart user will just stake from different wallets.

leviadam:
Looking at the protocol again it seems like there is a mismatch between the protocol and the code

PR #467, opened 2018-05-19
GenesisProtocol : Add RealMath library for accurate threshold calculations.

orenyodfat:


dev-matan-tsuberi:
This means all client code will need to change to always divide by 10^40 to get the actual result. Correct?

orenyodfat:
> This means all client code will need to change to always divide by 10^40 to get the actual result. Correct? 

No.  the return value is converted back -  RealMath.fromReal(res);

contracts/VotingMachines/GenesisProtocol.sol
```diff
@@ -538,13 +539,14 @@ contract GenesisProtocol is IntVoteInterface,UniversalScheme {
...
      */
     function threshold(bytes32 _proposalId,address _avatar) public view returns(int) {
-        uint e = 2;
+        int88 e = 2;
         Parameters memory params = parameters[proposals[_proposalId].paramsHash];
-        uint power = orgBoostedProposalsCnt[_avatar]/params.thresholdConstB;
-        if (power > 100) {
-            power = 100;
+        int128 power = RealMath.div(RealMath.toReal(int88(orgBoostedProposalsCnt[_avatar])),RealMath.toReal(int88(params.thresholdConstB)));
+        if (RealMath.fromReal(power) > 86) {
```
dkent600: Just curious:  Why has this changed fro 100 to 88?

contracts/VotingMachines/GenesisProtocol.sol
```diff
@@ -538,13 +539,14 @@ contract GenesisProtocol is IntVoteInterface,UniversalScheme {
      * @return int organization's score threshold.
      */
     function threshold(bytes32 _proposalId,address _avatar) public view returns(int) {
-        uint e = 2;
+        int88 e = 2;
         Parameters memory params = parameters[proposals[_proposalId].paramsHash];
-        uint power = orgBoostedProposalsCnt[_avatar]/params.thresholdConstB;
-        if (power > 100) {
```
dkent600: Just curious: Why has this changed from 100 to 86?

orenyodfat: The realMath lib works with int88 numbers

orenyodfat: Real math now works with int216 numbers so revert it back to 100 (which is huge enough number)

PR #475, opened 2018-06-11
GenesisProtocol - enable approve and stake in one transaction

orenyodfat:
Add _staker to stake function
enable approveandstake in one transaction
This is a lite version of #474

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -3,7 +3,7 @@ dist: trusty
 language: node_js
 
 node_js:
-  - "9.3.0"
+  - "9.10.0"
 
 before_install:
   - sudo apt-get update -qq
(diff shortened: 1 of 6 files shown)


dkent600:
Would like to see a similar API change for `ContributionReward.proposeContributionReward` and `VestingScheme.createVestedAgreement`

dkent600:
Would like to see a similar API change for `ContributionReward.proposeContributionReward` and `VestingScheme.createVestedAgreement`

orenyodfat:
@leviadam 
yes, you can stake(not vote) for someone else if he approve funds.
I guess we can have one api , if the msg.sender is the token so use the _staker other wise use msg.sender.
Though even that you can always stake for someone else using other token function(e.g transferAndCall or approveAndCall ...).
It is an issue!

PR #476, opened 2018-06-13
Add stakeWithSignature

orenyodfat:
One can approve and stake with one transaction by calling stakeWithSignature via the token contract ,which is ERC827 compatible.

A stake sign on a data blob which contain : 
 - genesisProtocol address   
 - proposalId
 - vote
 -amount
 -nonce - the nonce should be different for each call (per user) it  guaranty that the signature is unique and can be sent only one time to the contract.

Staker can stake by calling stakeWithSignature with proposalId,vote,amount,nonce and the signature(the signed data blob).
The contract side verify the params and fetch the staker address

### DIFF
--- a/contracts/VotingMachines/GenesisProtocol.sol
+++ b/contracts/VotingMachines/GenesisProtocol.sol
@@ -198,8 +198,10 @@ contract GenesisProtocol is IntVoteInterface,UniversalScheme {
         returns(bool)
         {
         require(stakeSignatures[_signature] == false);
-        bytes32 prefixedHash = prefixed(keccak256(address(this),_proposalId,_vote,_amount,_nonce));
+        // Builds a prefixed hash to mimic the behavior of eth_sign.
+        bytes32 prefixedHash = keccak256("\x19Ethereum Signed Message:\n32",keccak256(address(this),_proposalId,_vote,_amount,_nonce));
         address staker = prefixedHash.recover(_signature);
+        //a garbage staker address due to wrong signature will revert due to lack of approval and funds.
         require(staker!=address(0));
         stakeSignatures[_signature] = true;
         return _stake(_proposalId,_vote,_amount,staker);
@@ -828,11 +830,6 @@ contract GenesisProtocol is IntVoteInterface,UniversalScheme {
     function _isVotable(bytes32 _proposalId) private view returns(bool) {
         ProposalState pState = proposals[_proposalId].state;
         return ((pState == ProposalState.PreBoosted)||(pState == ProposalState.Boosted)||(pState == ProposalState.QuietEndingPeriod));
-    }
-
-    // Builds a prefixed hash to mimic the behavior of eth_sign.
-    function prefixed(bytes32 hash) private pure returns (bytes32) {
-        return keccak256("\x19Ethereum Signed Message:\n32", hash);
     }
 
     /**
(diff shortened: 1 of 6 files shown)


contracts/VotingMachines/GenesisProtocol.sol
```diff
@@ -175,42 +180,36 @@ contract GenesisProtocol is IntVoteInterface,UniversalScheme {
...
+        uint _vote,
+        uint _amount,
+        uint _nonce,
+        bytes _signature
+        )
+        external
+        returns(bool)
+        {
+        require(stakeSignatures[_signature] == false);
+        bytes32 prefixedHash = prefixed(keccak256(address(this),_proposalId,_vote,_amount,_nonce));
```
dev-matan-tsuberi: You can inline the function (`prefixed`) here, no need for it elsewhere.

contracts/VotingMachines/GenesisProtocol.sol
```diff
@@ -175,42 +180,36 @@ contract GenesisProtocol is IntVoteInterface,UniversalScheme {
...
+        address staker = prefixedHash.recover(_signature);
```
dev-matan-tsuberi: If the signature is invalid the address will be garbage, maybe we should require `staker` to be equal to `msg.sender` or `tx.origin` to prevent mistakes.

package.json
```diff
@@ -53,10 +53,10 @@
     "pug": "^2.0.0-rc.4",
     "rimraf": "^2.6.2",
     "run-with-ganache": "^0.1.1",
-    "solc": "^0.4.23",
+    "solc": "^0.4.24",
```
dev-matan-tsuberi: should update `pragma` on source files.

contracts/VotingMachines/GenesisProtocol.sol
orenyodfat: if the signature is invalid the return address is 0.

package.json
orenyodfat: Will update the pragma on other pr along fixing the warnings which popped up from 0.4.24.

contracts/VotingMachines/GenesisProtocol.sol
dev-matan-tsuberi: I think `ecrecover` may still return garbage if `r` or `s` are invalid.

orenyodfat: if the address is garbage the tx will revert .No approval and no funds..

contracts/VotingMachines/GenesisProtocol.sol
orenyodfat: done a936d94

contracts/VotingMachines/GenesisProtocol.sol
orenyodfat: comment added a936d94

dev-matan-tsuberi:
Should also be fixed in the `arc_private` repo.

package.json
orenyodfat: see pr #477

orenyodfat:
@dev-matan-tsuberi 
> Should also be fixed in the arc_private repo.

Done

dkent600:
What happens if the token is not ERC827 compatible, or is in fact evil?

orenyodfat:
@dkent600 

> What happens if the token is not ERC827 compatible, or is in fact evil?

As long as the token is ERC20/StandardToken compatible so staking can be done in a normal way.
If it is not supporting the functionality of approveAndCall (like ERC827 and other standards) so approve and stake in one transaction cannot be done.

If the token is evil,a dao decide to deploy  gp with an evil token , so no one will play the game with it...and it out of our scope.

Saying all that, GP is targeted to use GEN token which it ERC827. 
We are considering even to make the token address hard coded on the contract.

PR #478, opened 2018-06-18
GP : Threshold exclude expired boosted proposals

orenyodfat:
Currently expired boosted proposals are included in the score threshold calculation as it is not auto  executed on expiration.
In order to execute these proposals client need to actively do vote, stake or execute calls for these proposals.
This PR aim to solve it.
As it is not secure ,efficient and gas consuming to iterate over all boosted proposals ,the contract will  maintain a sorted array of expiration proposals times (which is relative cheap to maintain as elements are added only at the end of the array (O(1)). 
Calculating the expired proposals number can be done in O(LogN) where N is the number of boosted proposals.   
The number of  boosted proposals for the threshold calculation is done by subtracting the the expired proposals from the total boosted number of proposals.

### DIFF
--- a/contracts/VotingMachines/GenesisProtocol.sol
+++ b/contracts/VotingMachines/GenesisProtocol.sol
@@ -101,7 +101,7 @@ contract GenesisProtocol is IntVoteInterface,UniversalScheme {
     mapping(address=>uint) public orgBoostedProposalsCnt;
     StandardToken public stakingToken;
     mapping(address=>uint[]) public proposalsExpiredTimes; //proposals expired times
-    mapping(address=>uint) public quiteWindowProposals;
+    mapping(address=>uint) public quietWindowProposals;
     /**
      * @dev Constructor
      */
@@ -553,7 +553,7 @@ contract GenesisProtocol is IntVoteInterface,UniversalScheme {
           // solium-disable-next-line security/no-block-members
             expieredProposals = binarySearch(proposalsExpiredTimes[_avatar],0,proposalsExpiredTimes[_avatar].length-1,now);
         }
-        uint boostedProposals = orgBoostedProposalsCnt[_avatar].sub(expieredProposals).sub(quiteWindowProposals[_avatar]);
+        uint boostedProposals = orgBoostedProposalsCnt[_avatar].sub(expieredProposals).sub(quietWindowProposals[_avatar]);
         int216 e = 2;
 
         Parameters memory params = parameters[proposals[_proposalId].paramsHash];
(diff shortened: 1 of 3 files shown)


contracts/VotingMachines/GenesisProtocol.sol
```diff
@@ -100,6 +100,8 @@ contract GenesisProtocol is IntVoteInterface,UniversalScheme {
     uint public proposalsCnt; // Total number of proposals
     mapping(address=>uint) public orgBoostedProposalsCnt;
     StandardToken public stakingToken;
+    mapping(address=>uint[]) public proposalsExpiredTimes; //proposals expired times
+    mapping(address=>uint) public quiteWindowProposals;
```
dkent600: Should be "quietWindowProposals"

contracts/VotingMachines/GenesisProtocol.sol
```diff
@@ -100,6 +100,8 @@ contract GenesisProtocol is IntVoteInterface,UniversalScheme {
     uint public proposalsCnt; // Total number of proposals
     mapping(address=>uint) public orgBoostedProposalsCnt;
     StandardToken public stakingToken;
+    mapping(address=>uint[]) public proposalsExpiredTimes; //proposals expired times
```
dev-matan-tsuberi: Should also work when the DAO changes params (unregisters and registers). Maybe use a single [Heap](https://en.wikipedia.org/wiki/Heap_(data_structure)) per DAO.

contracts/VotingMachines/GenesisProtocol.sol
```diff
@@ -889,4 +901,21 @@ contract GenesisProtocol is IntVoteInterface,UniversalScheme {
         ProposalState pState = proposals[_proposalId].state;
         return ((pState == ProposalState.PreBoosted)||(pState == ProposalState.Boosted)||(pState == ProposalState.QuietEndingPeriod));
     }
+
+    function binarySearch(uint[] _data,uint _start,uint _end,uint _value) private view returns(uint) {
```
dev-matan-tsuberi: Need to check of `_data` gets pushed to the stack each time. If it is, there's a risk of stack overflow.

contracts/VotingMachines/GenesisProtocol.sol
```diff
@@ -545,9 +548,16 @@ contract GenesisProtocol is IntVoteInterface,UniversalScheme {
      * @return int organization's score threshold.
      */
     function threshold(bytes32 _proposalId,address _avatar) public view returns(int) {
+        uint expieredProposals;
```
dev-matan-tsuberi: typo

orenyodfat:
thanks @dev-matan-tsuberi for the input
As the gp params can be changed.. seems like a heap will be more appropriate as a data structure..for insert and search .

PR #481, opened 2018-06-19
Threshold exclude expired boosted proposals

orenyodfat:
Currently expired boosted proposals are included in the score threshold calculation as it is not auto executed on expiration.
In order to execute these proposals client need to actively do vote, stake or execute calls for these proposals.
This PR aim to solve it .
As it is not secure ,efficient and gas consuming to iterate over all boosted proposals ,the contract will maintain an [Order Statistic Tree ](https://en.wikipedia.org/wiki/Order_statistic_tree) with a balanced [AVL tree](https://en.wikipedia.org/wiki/AVL_tree) of expiration proposals times.
 - Inserting an element (proposal expiration time) to the tree is O(LogN) complexity.
 - Getting the number of the expired proposals for a given specific time (Rank) is O(LogN) complexity.

N  = the number of booster proposals.

In term of gas cost , inserting an element to the tree is relative more expensive than getting a Rank as its need to update the contract storage .
On current implementation insert  operation to a tree with 1000 elements (1000 boosted proposal) will cost on average around 400000 gas .It will not increase dramaticly with more elements due to the algorithm complexity O(LogN).
Getting a Rank from a tree with 1000 elements cost around 40000 gas.
Insert is done on a boosting event (on staking) and also when a proposal is shifting to a "quite window" state.
Rank is done for each threshold calculation.
All can be further optimise by doing kind of maintenance work and removing "old" expired boosted proposals from the tree by executing this proposals.
Please note that there is a protocol incentive for clients to execute expired positive proposals.
 
Remove an element from the tree has the same cost of insertion.

The number of boosted proposals for the threshold calculation is done by subtracting the the expired proposals from the total boosted number of proposals.

### DIFF
--- a/contracts/VotingMachines/GenesisProtocol.sol
+++ b/contracts/VotingMachines/GenesisProtocol.sol
@@ -5,6 +5,7 @@ import "./IntVoteInterface.sol";
 import "../universalSchemes/UniversalScheme.sol";
 import { RealMath } from "../libs/RealMath.sol";
 import "openzeppelin-solidity/contracts/ECRecovery.sol";
+import { OrderStatisticTree } from "../libs/OrderStatisticTree.sol";
 
 /**
  * @title GenesisProtocol implementation -an organization's voting machine scheme.
@@ -16,6 +17,7 @@ contract GenesisProtocol is IntVoteInterface,UniversalScheme {
     using RealMath for int216;
     using RealMath for int256;
     using ECRecovery for bytes32;
+    using OrderStatisticTree for OrderStatisticTree.Tree;
 
     enum ProposalState { None ,Closed, Executed, PreBoosted,Boosted,QuietEndingPeriod }
     enum ExecutionState { None, PreBoostedTimeOut, PreBoostedBarCrossed, BoostedTimeOut,BoostedBarCrossed }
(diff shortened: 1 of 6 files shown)


contracts/VotingMachines/GenesisProtocol.sol
```diff
@@ -83,6 +85,11 @@ contract GenesisProtocol is IntVoteInterface,UniversalScheme {
         mapping(address=>Staker) stakers;
     }
 
+    struct ExpieredProposal {
```
ben-kaufman: This struct seems to never be used. Does it have any future purpose? If not, I guess it should be removed.

test/orderstatistictree.js
```diff
@@ -0,0 +1,76 @@
...
+
+    var rank;
+    assert.equal(count,values.length);
+    for (i = 0 ;i< count ;i++) {
+      rank = await orderStatisticTree.rank(values[i]);
+      assert.equal(ranks[i],rank);
+    }
+
+    var valuesInBetween = [4,8,20,101,2];
+    var ranksInBetween  = [2,5,5,7,1];
```
ben-kaufman: I recommend adding to the test a duplicate case just to cover that up as well. Maybe rank 5 or 6 to see make sure the ranking supports duplicates as well.

contracts/VotingMachines/GenesisProtocol.sol
orenyodfat: Removed at c9c7256

test/orderstatistictree.js
orenyodfat: The duplicate case is already covered here https://github.com/daostack/arc/blob/threshold_exclude_expiered_proposals/test/orderstatistictree.js#L11 .
Did I missed something ?

ben-kaufman: No, sorry, I missed that one.

PR #490, opened 2018-07-09
Locking4reputation

orenyodfat:
A schemes for locking ETH and Token for reputation allocation.


contracts/schemes/Locking4Reputation.sol
```diff
@@ -0,0 +1,127 @@
...
+        // solium-disable-next-line security/no-block-members
+        require(block.timestamp >= lockingEndTime,"check the lock period pass");
+        require(scores[_beneficiary] > 0,"score should be > 0");
+        uint score = scores[_beneficiary];
+        scores[_beneficiary] = 0;
+        int256 repRelation = int216(score).toReal().mul(int216(reputationReward).toReal());
+        uint reputation = uint256(repRelation.div(int216(totalScore).toReal()).fromReal());
+        require(ControllerInterface(avatar.owner()).mintReputation(reputation,_beneficiary,avatar));
+        //check that the reputation is sum zero
+        reputationRewardLeft = reputationRewardLeft.sub(reputation);
```
leviadam: If you want this to be a better protection I would put it before line 74.

contracts/schemes/LockingToken4Reputation.sol
```diff
@@ -0,0 +1,54 @@
...
+                       _lockingEndTime,
+                       _maxLockingPeriod)
+    public
+    {
+        token = _token;
+    }
+
+    function release(address _beneficiary,bytes32 _lockingId) public returns(bool) {
+        uint amount = super._release(_beneficiary,_lockingId);
+        return token.transfer(_beneficiary,amount);
```
leviadam: Maybe put `require` or use `safeTransfer`

PR #508, opened 2018-07-30
GenesisProtocol -threshold function receive paramHash (instead of proposalId)

### LINKED ISSUE #500: Don't require passing in a proposalId to GenesisProtocol.threshold
The current threshold should be the same for the whole DAO, not per proposal. Can't we track that param by the avatar so I don't have to look up the new value for every proposal after every stake?

orenyodfat:
resolve #500 

Fix warning from IntVoteIntraface

### DIFF
--- a/contracts/VotingMachines/AbsoluteVote.sol
+++ b/contracts/VotingMachines/AbsoluteVote.sol
@@ -243,23 +243,8 @@ contract AbsoluteVote is IntVoteInterface {
       * @return bool true - the proposal has been executed
       *              false - otherwise.
      */
-    function execute(bytes32 _proposalId) public votable(_proposalId) returns(bool) {
-        Proposal storage proposal = proposals[_proposalId];
-        Reputation reputation = parameters[proposal.paramsHash].reputationSystem;
-        require(reputation != address(0));
-        uint totalReputation = reputation.totalSupply();
-        uint precReq = parameters[proposal.paramsHash].precReq;
-        // Check if someone crossed the bar:
-        for (uint cnt = 0; cnt <= proposal.numOfChoices; cnt++) {
-            if (proposal.votes[cnt] > totalReputation*precReq/100) {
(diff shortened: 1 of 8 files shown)


contracts/VotingMachines/IntVoteInterface.sol
```diff
@@ -34,26 +34,26 @@ interface IntVoteInterface {
         ) external returns(bytes32);
 
     // Only owned proposals and only the owner:
-    function cancelProposal(bytes32 _proposalId) external onlyProposalOwner(_proposalId) votable(_proposalId) returns(bool);
```
dkent600: What are the use cases that motivate removing `onlyProposalOwner` and `votable` in all these cases?

contracts/VotingMachines/AbsoluteVote.sol
```diff
@@ -300,6 +285,31 @@ contract AbsoluteVote is IntVoteInterface {
...
     }
 
+    /**
+      * @dev execute check if the proposal has been decided, and if so, execute the proposal
+      * @param _proposalId the id of the proposal
+      * @return bool true - the proposal has been executed
+      *              false - otherwise.
+     */
+    function _execute(bytes32 _proposalId) internal votable(_proposalId) returns(bool) {
+        Proposal storage proposal = proposals[_proposalId];
```
dev-matan-tsuberi: Why this change?

orenyodfat:
execute function should be external as it appears in the interface though
in the other end it suppose to be called internally .
This modification is to make the compiler happy as it was complaining about
that .

orenyodfat:
Empty modifiers are ignored .
The compiler was complaining about it .

contracts/VotingMachines/IntVoteInterface.sol
orenyodfat: Empty modifiers are ignored .
The compiler was complaining about it .

PR #534, opened 2018-09-12
locking schems : add setParameters (one time call)

orenyodfat:
Add one time ,onlyOwner , initialiser (setParameters) to ease the deployment processes for the following schemes: 

- Auction4Reputation
- ExternalLocking4Reputation
- FixReputationAllocation
- LockingEth4Reputation
- LockingToken4Reputation


contracts/schemes/Auction4Reputation.sol
```diff
@@ -38,7 +39,7 @@ contract Auction4Reputation {
     address public wallet;
 
     /**
-     * @dev constructor
+     * @dev setParameters
```
leviadam: The name `setParameters` is an abuse of notation, as we use it in Arc for a different concept.
I think it is better to change it to `initialize` or something similar.

orenyodfat: done at f98d948

PR #550, opened 2018-10-07
FixedReputationAllocation get redeemEnableTime as initialise param

### LINKED ISSUE #549: Remove lockingId from Locking4Reputation.redeem and the Redeem event
`Locking4Reputation.redeem` takes `_lockingId` for an argument but only uses it for the `Redeem` event that is emitted.  However, a lockingId is completely irrelevant to the redemption, as `redeem` pays out the entire reputation and score for all locks that the beneficiary has created.

The presence of the lockingId, and the code comment about it, imply that the redemption is only paid for the amount locked.  If this is true, then the current behavior of `redeem` is wrong.

orenyodfat:
resolve #549

### DIFF
--- a/contracts/schemes/FixReputationAllocation.sol
+++ b/contracts/schemes/FixReputationAllocation.sol
@@ -26,16 +26,19 @@ contract FixedReputationAllocation is Ownable {
     bool public isEnable;
     uint public numberOfBeneficiaries;
     uint public beneficiaryReward;
+    uint public redeemEnableTime;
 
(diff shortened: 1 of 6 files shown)


orenyodfat:
Add beneficiaries can be done only if not enable..once enable the "white list" of beneficiaries cannot be changed.
https://github.com/daostack/arc/blob/gh-issue-549/contracts/schemes/FixReputationAllocation.sol#L67

PR #552, opened 2018-10-08
Use infra 08 

orenyodfat:
use solc v 0.4.25

bump arc v to 54

### DIFF
--- a/contracts/VotingMachines/VotingMachineCallbacks.sol
+++ b/contracts/VotingMachines/VotingMachineCallbacks.sol
@@ -1,4 +1,4 @@
-pragma solidity ^0.4.24;
+pragma solidity ^0.4.25;
 
 import "../universalSchemes/UniversalScheme.sol";
 import "@daostack/infra/contracts/VotingMachines/GenesisProtocol.sol";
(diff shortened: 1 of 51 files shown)


contracts/VotingMachines/VotingMachineCallbacks.sol
```diff
@@ -67,4 +67,13 @@ contract GenesisProtocolCallbacks is GenesisProtocolCallbacksInterface {
         return ControllerInterface(avatar.owner()).externalTokenTransfer(_stakingToken,_beneficiary,_amount,address(avatar));
     }
 
+    function balanceOfStakingToken(StandardToken _stakingToken,bytes32 _proposalId) external view returns(uint) {
+        Avatar avatar = proposalsInfo[_proposalId].avatar;
+        if (avatar == Avatar(0)) {
+            return 0;
+        }
+        _stakingToken.balanceOf(address(avatar));
```
leviadam: Shouldn't there be a `return` here?
Also, isn't the `if` redundant?

orenyodfat: fixed at 9834bf0

PR #558, opened 2018-10-15
add forwarder scheme

orenyodfat:
This scheme behave like a "buffer" between 2 daos.
1. It has an expiration time - after this time it's forwarding function is disable.
2. It forward any call from one dao to other dao controller.

### DIFF
--- a/contracts/schemes/Forwarder.sol
+++ b/contracts/schemes/Forwarder.sol
@@ -40,6 +40,8 @@ contract Forwarder is Ownable {
      * @param _expirationTime the expirationTime to forwardCall
      */
     function initialize(Avatar _avatar, uint _expirationTime) external onlyOwner {
+        require(avatar == Avatar(0), "can be called only one time");
+        require(_avatar != Avatar(0), "avatar cannot be zero");
         avatar = _avatar;
         expirationTime = _expirationTime;
     }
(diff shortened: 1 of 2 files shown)


contracts/schemes/Forwarder.sol
```diff
@@ -0,0 +1,56 @@
...
+        default { return(0, returndatasize) }
+        }
+    }
+
+    /**
+     * @dev initialize
+     * @param _avatar the avatar of the dao to forward the call to
+     * @param _expirationTime the expirationTime to forwardCall
+     */
+    function initialize(Avatar _avatar, uint _expirationTime) external onlyOwner {
```
leviadam: Here there is basically no "time bomb" as the owner can change it.

contracts/schemes/Forwarder.sol
```diff
@@ -0,0 +1,56 @@
...
+
+/**
+ * @title A scheme to forward a call to a dao.
+ *        The scheme can unregister itself when its expirationTime reached.
+ */
+
+
+contract Forwarder is Ownable {
+
+    Avatar public avatar;
```
leviadam: Why Avatar?
Why not make generic and initialize some address.

orenyodfat: the avatar is needed to call for unregisterSelf .
we can add another address to make it more generic though I am not sure it is needed.

contracts/schemes/Forwarder.sol
orenyodfat: fix 51552e6

contracts/schemes/Forwarder.sol
leviadam: Not needed. This is OK.

PR #562, opened 2018-11-18
ExternalLocking4Rep : add registrar

orenyodfat:
- rename lock  to claim .

- claiming can be done for behalf of someone else only if it is register .

- add register function.

### DIFF
--- a/contracts/schemes/ExternalLocking4Reputation.sol
+++ b/contracts/schemes/ExternalLocking4Reputation.sol
@@ -10,20 +10,24 @@ import "openzeppelin-solidity/contracts/ownership/Ownable.sol";
 
 contract ExternalLocking4Reputation is Locking4Reputation, Ownable {
 
+    event Register(address indexed _beneficiary);
+
     address public externalLockingContract;
     string public getBalanceFuncSignature;
 
     // locker -> bool
     mapping(address => bool) public externalLockers;
+    //      beneficiary -> bool
+    mapping(address     => bool) public registrar;
 
     /**
      * @dev initialize
      * @param _avatar the avatar to mint reputation from
      * @param _reputationReward the total reputation this contract will reward
      *        for the token locking
-     * @param _lockingStartTime locking starting period time.
-     * @param _lockingEndTime the locking end time.
-     *        locking is disable after this time.
+     * @param _claimingStartTime claiming starting period time.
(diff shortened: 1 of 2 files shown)


contracts/schemes/ExternalLocking4Reputation.sol
```diff
@@ -10,20 +10,24 @@ import "openzeppelin-solidity/contracts/ownership/Ownable.sol";
 
 contract ExternalLocking4Reputation is Locking4Reputation, Ownable {
 
+    event Register(address indexed _beneficiary);
+
     address public externalLockingContract;
     string public getBalanceFuncSignature;
 
     // locker -> bool
     mapping(address => bool) public externalLockers;
```
dkent600: I think the code would be more clear if this were called "claimers".

contracts/schemes/ExternalLocking4Reputation.sol
```diff
@@ -33,36 +37,45 @@ contract ExternalLocking4Reputation is Locking4Reputation, Ownable {
...
     /**
-     * @dev lock function
-     * @return lockingId
+     * @dev claim function
+     * @param _beneficiary the beneficiary address to claim for
+     *        if _beneficiary == 0 the claim will be for the msg.sender.
+     * @return claimId
      */
-    function lock() public returns(bytes32) {
+    function claim(address _beneficiary) public returns(bytes32) {
```
dkent600: Can we call "_beneficiary"  "_claimer"?

leviadam: But it is not the claimer, as one can claim for someone else, who was registered.

contracts/schemes/ExternalLocking4Reputation.sol
dkent600: Even better:  "claimants".

contracts/schemes/ExternalLocking4Reputation.sol
dkent600: @leviadam 

msg.sender in this case is claiming on behalf of the claimant (here `_beneficiary`), who "registered"  (poor word IMHO) to allow someone else (here msg.sender) to "claim" on their behalf. 

`msg.sender` is not here claiming anything, rather is executing a claim an behalf of someone else.

PR #565, opened 2018-11-28
commit migrated_contracts

bohendo:
arc.js contains an extremely valuable folder: `migrated_contracts`.

This is a folder that 3rd party developers can use as a central source of truth for abis and, more importantly, addresses of arc contracts that have been deployed to production aka mainnet.

I feel strongly that this folder should not be part of the client-side library as it's strongly associated with the contracts themselves.

As a 3rd party dev, I'd like to access these contract artifacts via this arc repo so that I can avoid depending on such a bulky package (arc.js) in situations where the only thing I need is a canonical reference to these contract addresses.

If the future, as the next generation of contracts are deployed, I'd like to see these contract artifacts accumulate in this repo, maybe following a pattern that looks something like:

```
migrated_contracts # or maybe contract_artifacts if we follow truffle's naming convention
  - v1
     - GenesisProtocol.json
     - ContributionReward.json
  - v2
     - GenesisProtocol.json
     - ContributionReward.json
```


orenyodfat:
@bohendo we already moved the migration out to a separate repo : https://github.com/daostack/migration

bohendo:
I think other devs in the ecosystem would appreciate having a canonical reference to the mainnet contract addresses without needing to depend on arc.js.

The `@daostack/migrations` module doesn't seem to provide this.

orenyodfat:
The @daostack/migrations does not depend on arc.js

bohendo:
@orenyodfat That's great but `@daostack/migrations` also doesn't include `migrated_contracts` which is a very useful folder & is something I'd like to have easier access to

dkent600:
@bohendo 

Arc already does include these json files in its package:

See: node_modules\@daostack\arc\build\contracts

bohendo:
The artifacts in `node_modules/@daostack/arc/build/contracts` don't contain any addresses of deployed contracts.

As far as I can tell, these addresses only exist in `@daostack/arc.js/migrated_contracts` which is unfortunate for applications that only need these addresses & contract artifacts but not any thing else. 

Arc.js is an extremely bulky library and it's inconvenient to have to download it when 99% of it's contents aren't needed.

PR #569, opened 2018-12-08
redeem do nothing, return zero when no score/bid

dkent600:
It is a pretty big pain to code around that the reputation bootstrapping contracts `redeem` function reverts when score or bid are > zero.  Is there a realy need to revert in that case?

This PR changes the behavior to do nothing and return 0 when redeem is attempted with no score or bid.


orenyodfat:
One should check if the redeem value before redeem .
Use call

dkent600:
Using 'call' doesn't help -- it reverts just like when not using it` , and one must then determine *why* it reverted (because it can revert for other reasons).  Better is to check in advance whether the account has a score (or a bid).  But is there any reason to force this?  The revert in this case seems unnecessary.

orenyodfat:
Just do call and check the revert reasone if it is revert

PR #590, opened 2019-01-16
Add metaData to avatar

### LINKED ISSUE #589: Add metadata to a DAO
It seems very reasonable that a DAO would want to have metadata for the DAO. Things like manifesto, roadmap, description, maybe even color-palette in the future.
One can see this coming from the genesis alpha community, as something of the sort was suggested in a  [PR](https://github.com/daostack/alchemy/pull/215) in alchemy.

I suggest the avatar should have a metadata function, where all it is doing is emiting an event with the hash of an IPFS json file that will contain the metadata.

leviadam:
This PR resolves #589.

### DIFF
--- a/contracts/controller/Avatar.sol
+++ b/contracts/controller/Avatar.sol
@@ -23,6 +23,7 @@ contract Avatar is Ownable {
     event ExternalTokenTransferFrom(address indexed _externalToken, address _from, address _to, uint256 _value);
     event ExternalTokenApproval(address indexed _externalToken, address _spender, uint256 _value);
     event ReceiveEther(address indexed _sender, uint256 _value);
+    event MetaData(string _metaData);
 
     /**
     * @dev the constructor takes organization name, native token and reputation system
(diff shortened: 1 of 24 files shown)


contracts/controller/Avatar.sol
```diff
@@ -121,4 +122,15 @@ contract Avatar is Ownable {
         return true;
     }
 
+    /**
+    * @dev metaData emits an event with a string, should contain the hash of some meta data.
+    * @param _metaData a string representing a hash of the meta data
+    * @return bool which represents a success
+    */
+    function metaData(string memory _metaData) public onlyOwner returns(bool) {
```
orenyodfat: this is onlyOwner  ..
should we add this api to the controller interface ?

leviadam: Probably yes.
For me it looked like the important thing is to have it on the avatar, as the controller is meant to be upgradable.
But we can decide to support it already in the controller as well.

orenyodfat: i guess it should also be under certain permission...

leviadam: Also for the time being an org can just put something there on deployment, and only later change it with when we support this feature.

orenyodfat: you mean via a param to  the DaoCreator ?

leviadam: The permissions issue is one of the reasons I thought to avoid this for the time being. It may be also under the upgrade permission.

leviadam: Yes, regarding the DaoCreator.

leviadam:
Looks good.

Issue #539, opened 2018-09-25
Locking4Reputation totalLocked and totalLockedLeft can be more clear


dkent600:
It is not clear from the code what the `Locking4Reputation` `totalLocked` and `totalLockedLeft` properties represent.  I'm not sure the code is correct by any interpretation.

After `_lock`, totalLocked is incremented and then totalLockedLeft = totalLocked .

After `_release`, totalLocked is unchanged while totalLockedLeft is decremented by the released amount.

Can we at least document and if needed, fix these properties?

dkent600:
@orenyodfat

orenyodfat:
This is done in order to add addition check and verify there is sum zero of the totalLocked .--> cannot redeem more than locked.

dkent600:
Suppose:

1. lock 10
totalLockedLeft  = totalLocked  = 10
2. lock 10
totalLockedLeft  = totalLocked  = 20
3. release lockid 1
totalLockedLeft  = 10
totalLocked  = 20
4. lock 10
totalLockedLeft  = totalLocked  = 30

Does it make sense that total totalLockedLeft  is 30 after releasing in step 4?

dkent600:
OK, it makes sense when one realises that one can only lock during the locking period and only redeem once the period has expired.

PR #598, opened 2019-02-14
fix totalLockedLeft calculation

orenyodfat:
locking4reputation scheme


dkent600:
Looks like this Resolves:  https://github.com/daostack/arc/issues/539

PR #599, opened 2019-02-16
add _auctionId to bid function parameteres

### LINKED ISSUE #594: Should not be able to bid to the wrong auction
**STEPS**

1. An application informs the user that a certain auction is in progress.
2. The user places a bid, expecting it to be placed on this auction
3. Before execution reaches the Auction4Reputation contract, the blockchain time advances such that the auction number also advances.  (Easy to happen if the user delays responding to MM prompts)

**RESULT**

The bid is not placed on the auction that the user intended.  User is surprised and disappointed.

**NOTES**

There is no way to completely avoid the possibility of voting on the wrong auction unless we change the contract's bid method method to take an additional parameter auctionId which would be the id of the auction in which we intend to bid. The contract bid method would revert if the current auctionId is != to the one we intend to bid in.

orenyodfat:
fix #594

### DIFF
--- a/contracts/schemes/Auction4Reputation.sol
+++ b/contracts/schemes/Auction4Reputation.sol
@@ -122,7 +122,7 @@ contract Auction4Reputation is Ownable {
         address(token).safeTransferFrom(msg.sender, address(this), _amount);
         // solhint-disable-next-line not-rely-on-time
         auctionId = (now - auctionsStartTime) / auctionPeriod;
-        require(auctionId == _auctionId, "false auction id");
+        require(auctionId == _auctionId, "auction is not active");
         Auction storage auction = auctions[auctionId];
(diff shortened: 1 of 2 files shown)


contracts/schemes/Auction4Reputation.sol
```diff
@@ -121,6 +122,7 @@ contract Auction4Reputation is Ownable {
         address(token).safeTransferFrom(msg.sender, address(this), _amount);
         // solhint-disable-next-line not-rely-on-time
         auctionId = (now - auctionsStartTime) / auctionPeriod;
+        require(auctionId == _auctionId, "false auction id");
```
dkent600: More clear revert string would be "auction has completed"

Or if that presumes too much, then "auction is not active"

PR #606, opened 2019-03-12
Add value (ETH) to genericCall

orenyodfat:


contracts/controller/Avatar.sol
```diff
@@ -17,7 +17,7 @@ contract Avatar is Ownable {
     DAOToken public nativeToken;
     Reputation public nativeReputation;
 
-    event GenericCall(address indexed _contract, bytes _params, bool _success);
+    event GenericCall(address indexed _contract, bytes _params, bool _success, uint _value);
```
leviadam: I think the order of the parameters in the event should be different, and params should also be renamed to _data:
`address indexed _contract, bytes _data, uint _value, bool _success`

orenyodfat: see 
af108f1

PR #610, opened 2019-03-14
ContributionReward : Remove proposing fee

orenyodfat:
This is a breaking change.


dkent600:
Maybe a couple of use cases for DAOs that might want to charge a fee:

1. My DAO gives very high rewards for winning proposals, and so wants agents to pay for the opportunity to win those rewards
2. My DAO has a very high overhead cost of processing proposals, and so wants to recoup some or all of that cost.

orenyodfat:
> Maybe a couple of use cases for DAOs that might want to charge a fee:
> 
> 1. My DAO gives very high rewards for winning proposals, and so wants agents to pay for the opportunity to win those rewards
> 2. My DAO has a very high overhead cost of processing proposals, and so wants to recoup some or all of that cost.

Not sure about these , though any how I guess it should be part of the protocol (not in this place).

PR #613, opened 2019-03-23
VestingScheme + uController fixes

### LINKED ISSUE #608: VestingScheme : potential vulnerability
Avatar and Controller contracts are incorrectly expected to follow native behaviour when nativeToken() and mintTokens() is called.

orenyodfat:
VestingScheme: Check tokens actually minted. fix #608

uController : consolidate reputations and tokens mapping to prevent switching between token and reputation.

externalLocking4Reputation: replace add(0x20,0) with 0x20.

### DIFF
--- a/contracts/controller/UController.sol
+++ b/contracts/controller/UController.sol
@@ -47,7 +47,6 @@ contract UController is ControllerInterface {
         mapping(address=>GlobalConstraintRegister) globalConstraintsRegisterPre;
       // globalConstraintsRegisterPost indicate if a globalConstraints is registered as a Post global constraint.
         mapping(address=>GlobalConstraintRegister) globalConstraintsRegisterPost;
-        bool exist;
     }
 
     //mapping between organization's avatar address to Organization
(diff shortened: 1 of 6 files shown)


contracts/controller/UController.sol
```diff
@@ -98,14 +94,14 @@ contract UController is ControllerInterface {
         require(nativeToken.owner() == address(this));
         require(nativeReputation.owner() == address(this));
         //To guaranty uniqueness for the reputation systems.
-        require(!reputations[address(nativeReputation)]);
+        require(!actors[address(nativeReputation)]);
+        actors[address(nativeReputation)] = true;
         //To guaranty uniqueness for the nativeToken.
-        require(!tokens[address(nativeToken)]);
+        require(!actors[address(nativeToken)]);
+        actors[address(nativeToken)] = true;
```
leviadam: Should we not do the same for the avatar?

orenyodfat: see e5c68db

PR #629, opened 2019-04-23
Add agreement hash to locking schemes

leviadam:


contracts/schemes/Auction4Reputation.sol
```diff
@@ -80,6 +83,7 @@ contract Auction4Reputation {
         auctionReputationReward = _auctionReputationReward;
         reputationRewardLeft = _auctionReputationReward.mul(_numberOfAuctions);
         redeemEnableTime = _redeemEnableTime;
+        agreementHash = _agreementHash;
```
dkent600: Would it be helpful, for each contract, to require that the hash be != 0x0?

A 0x0 hash would not fulfill the purpose of the feature.

leviadam: I think it is good yo have 0, for the people who dont want this feature.

dkent600: OK, I have previous heard the opposite argument for not undertaking other suggestions I've made about the bootstrapping contracts:  That no one else is ever going to use these contracts.  lol

orenyodfat:
how about adding AgreementHash contract  ?
 the locking contracts can inherit from this contract.
the contract will provide a modifier `onlyAgree()`

leviadam:
LGTM

PR #631, opened 2019-05-01
add ReputationFromToken contract

orenyodfat:
this scheme enable users to mint(redeem) reputation according to specific erc20 token holding.

### DIFF
--- a/contracts/schemes/ReputationFromToken.sol
+++ b/contracts/schemes/ReputationFromToken.sol
@@ -1,0 +1,53 @@
+pragma solidity ^0.5.4;
+
+import "../Controller/ControllerInterface.sol";
+/**
+ * @title A scheme for reputation allocation according to token balances
+ */
+
+contract ReputationFromToken {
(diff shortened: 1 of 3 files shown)


dkent600:
Would be nice to describe in the PR description why this contract is needed, what is its value?

contracts/schemes/ReputationFromToken.sol
```diff
@@ -0,0 +1,53 @@
...
+
+    /**
+     * @dev redeem function
+     * @param _beneficiary the beneficiary address to redeem for
+     */
+    function redeem(address _beneficiary) public {
+        require(avatar != Avatar(0), "should initialize first");
+        require(redeems[msg.sender] == false, "redeeming twice from the same account is not allowed");
+        redeems[msg.sender] = true;
+        (bool result, bytes memory returnValue) =
```
leviadam: Isn't it better to use the ERC20 interface?

PR #632, opened 2019-05-02
Handle _beneficiary=0 in rep from token

leviadam:


contracts/schemes/ReputationFromToken.sol
```diff
@@ -38,6 +39,9 @@ contract ReputationFromToken {
         require(redeems[msg.sender] == false, "redeeming twice from the same account is not allowed");
         redeems[msg.sender] = true;
         uint256 tokenAmount = tokenContract.balanceOf(msg.sender);
+        if (_beneficiary == address(0)) {
+            _beneficiary = msg.sender;
```
orenyodfat: why is this needed ?
if it is needed , please add test.

leviadam: 0 is default value, we should handle it if it is called by mistake, and also allow call with no value.

orenyodfat: no issue. 
though other fault values can be submitted as well.

test/reputationfromtoken.js
```diff
@@ -56,6 +56,18 @@ contract('ReputationFromToken', accounts => {
...
+    it("redeem with no beneficiary", async () => {
+      let testSetup = await setup(accounts);
+      var tx = await testSetup.reputationFromToken.redeem(helpers.NULL_ADDRESS);
+      assert.equal(tx.logs.length,1);
+      assert.equal(tx.logs[0].event,"Redeem");
+      assert.equal(tx.logs[0].args._beneficiary,accounts[0]);
+      assert.equal(tx.logs[0].args._amount,100);
+      assert.equal(tx.logs[0].args._sender,accounts[0]);
+      assert.equal(await testSetup.org.reputation.balanceOf(accounts[0]),1100);
+      assert.equal(await testSetup.org.reputation.balanceOf(accounts[1]),0);
```
orenyodfat: does checking accounts[1] balance is needed ?

leviadam: Not really, doesn't hurt.

PR #640, opened 2019-05-31
DAOTracker: A Solution For Indexing New DAOs

dOrgJelli:
**[Problem]**
As discussed [in this thread](https://github.com/daostack/subgraph/issues/197), we need a way to index newly deployed DAOs that satisfies the following constraints:
  1. Protects against DoS / spam.  
      * Reference: https://github.com/daostack/subgraph/issues/197#issuecomment-489294562  
  2. Doesn't require new DAOs to propose themselves to a single "central" `DAORegistry`.
      * Reference: https://github.com/daostack/subgraph/issues/197#issuecomment-489294562  
  3. Doesn't require the use of the `DaoCreator` contract or the `UController`.  
      * Reference: https://github.com/daostack/subgraph/issues/197#issuecomment-487826482

**[Solution]**
Create a new on-chain "source of truth" that dictates what the subgraph is indexing. All DAOs that wish to be cached will need to call the `track(avatar, controller)` method on this contract *immediately after they've been created* to ensure no events will be lost. This should typically happen after Avatar, DAOToken, Reputation, and Controller have all been deployed, and no other setup actions have taken place.

**[Question]**
Rename `DAOTracker` to `DAONetwork`?

**[Contract Explained]**
* `track(avatar, controller)`: [read description commented here](https://github.com/dOrgTech/arc/blob/0d5ab2686c76e646030c2a23a2ce4872c5b88f82/contracts/utils/DAOTracker.sol#L37)
* `blacklist(avatar)`: [read description commented here](https://github.com/dOrgTech/arc/blob/0d5ab2686c76e646030c2a23a2ce4872c5b88f82/contracts/utils/DAOTracker.sol#L67)
* `reset(avatar)`: [read description commented here](https://github.com/dOrgTech/arc/blob/0d5ab2686c76e646030c2a23a2ce4872c5b88f82/contracts/utils/DAOTracker.sol#L80)
* owner: Currently since DAOstack is operating the caching server, they would be the owner of this contract. In the future this could be a collective effort.

**[This is part 1 of a 3 part update]**
  1. Arc: This PR.
  2. Migration: Fix Migration scripts as this will be a breaking change (DaoCreator constructor).
  3. Subgraph: Add this contract to the subgraph, and add the following functionality:
```typescript
export function handleTrackDAO(event: TrackDAO) {
  // Check to see if the DAO is already in the cache
  if (store.get("DAO", event.args._avatar) != null) {
    return;
  }

  // Check to see if the DAO being tracked is blacklisted
  // NOTE: the event we're processing could be from any historic block (ex: 5),
  //    but this smart contract view method is going to return the latest block's state.
  //    This detail needs to be verified, as the method could return the current block's state.
  const daoTrackerSC = DAOTracker.bind(event.address);
  const trackingInfo = daoTrackerSC.methods.tracked(event.args._avatar);

  if (trackingInfo.blacklist) {
    return;
  }

  // Start tracking the newly added DAO
  AvatarTemplate.create(event.args._avatar);
  DAOTokenTemplate.create(trackingInfo.nativeToken);
  ReputationTemplate.create(trackingInfo.nativeReputation);

  // Make sure the controller isn't the Universal Controller
  if (store.get("UniversalController", trackingInfo.controller) == null) {
    ControllerTemplate.create(trackingInfo.controller);
  }
}
```

### DIFF
--- a/contracts/universalSchemes/DaoCreator.sol
+++ b/contracts/universalSchemes/DaoCreator.sol
@@ -3,6 +3,7 @@ pragma solidity ^0.5.4;
 import "./UniversalScheme.sol";
 import "../controller/UController.sol";
 import "../controller/Controller.sol";
+import "../utils/DAOTracker.sol";
 
 
 /**
@@ -29,9 +30,13 @@ contract DaoCreator {
     event InitialSchemesSet (address _avatar);
 
     ControllerCreator private controllerCreator;
+    DAOTracker private daoTracker;
 
-    constructor(ControllerCreator _controllerCreator) public {
+    constructor(ControllerCreator _controllerCreator, DAOTracker _daoTracker) public {
+        require(_controllerCreator != ControllerCreator(0));
+        require(_daoTracker != DAOTracker(0));
         controllerCreator = _controllerCreator;
+        daoTracker = _daoTracker;
     }
 
     /**
@@ -169,8 +174,7 @@ contract DaoCreator {
         uint[] memory _foundersTokenAmount,
         uint[] memory _foundersReputationAmount,
         UController _uController,
-        uint256 _cap,
-        DAOTracker _daoTracker
+        uint256 _cap
     ) private returns(address)
     {
         // Create Token, Reputation and Avatar:
@@ -201,8 +205,7 @@ contract DaoCreator {
         }
 
         // Add the DAO to the tracking registry
-        require(_daoTracker != address(0));
-        _daoTracker.track(avatar, controller);
+        daoTracker.track(avatar, controller);
 
         // Transfer ownership:
         avatar.transferOwnership(address(controller));
--- a/contracts/universalSchemes/DaoCreator.sol
+++ b/contracts/universalSchemes/DaoCreator.sol
@@ -8,7 +8,6 @@ import "../controller/Controller.sol";
 /**
  * @title ControllerCreator for creating a single controller.
  */
-
 contract ControllerCreator {
 
     function create(Avatar _avatar) public returns(address) {
@@ -22,8 +21,6 @@ contract ControllerCreator {
 /**
  * @title Genesis Scheme that creates organizations
  */
-
-
 contract DaoCreator {
 
     mapping(address=>address) public locks;
@@ -32,9 +29,11 @@ contract DaoCreator {
     event InitialSchemesSet (address _avatar);
 
     ControllerCreator private controllerCreator;
+    DAOTracker private daoTracker;
 
-    constructor(ControllerCreator _controllerCreator) public {
+    constructor(ControllerCreator _controllerCreator, DAOTracker _daoTracker) public {
         controllerCreator = _controllerCreator;
+        daoTracker = _daoTracker;
     }
 
     /**
@@ -198,16 +197,19 @@ contract DaoCreator {
         // Create Controller:
         if (UController(0) == _uController) {
             controller = ControllerInterface(controllerCreator.create(avatar));
-            avatar.transferOwnership(address(controller));
-            // Transfer ownership:
-            nativeToken.transferOwnership(address(controller));
-            nativeReputation.transferOwnership(address(controller));
         } else {
             controller = _uController;
-            avatar.transferOwnership(address(controller));
-            // Transfer ownership:
-            nativeToken.transferOwnership(address(controller));
-            nativeReputation.transferOwnership(address(controller));
+        }
+
+        // Add the DAO to the tracking registry
+        daoTracker.track(avatar, controller);
+
+        // Transfer ownership:
+        avatar.transferOwnership(address(controller));
+        nativeToken.transferOwnership(address(controller));
+        nativeReputation.transferOwnership(address(controller));
+
+        if (controller == _uController) {
             _uController.newOrganization(avatar);
         }
 
--- a/contracts/universalSchemes/DaoCreator.sol
+++ b/contracts/universalSchemes/DaoCreator.sol
@@ -29,11 +29,9 @@ contract DaoCreator {
     event InitialSchemesSet (address _avatar);
 
     ControllerCreator private controllerCreator;
-    DAOTracker private daoTracker;
 
-    constructor(ControllerCreator _controllerCreator, DAOTracker _daoTracker) public {
+    constructor(ControllerCreator _controllerCreator) public {
         controllerCreator = _controllerCreator;
-        daoTracker = _daoTracker;
     }
 
     /**
@@ -171,7 +169,8 @@ contract DaoCreator {
         uint[] memory _foundersTokenAmount,
         uint[] memory _foundersReputationAmount,
         UController _uController,
-        uint256 _cap
+        uint256 _cap,
+        DAOTracker _daoTracker
     ) private returns(address)
     {
         // Create Token, Reputation and Avatar:
@@ -202,7 +201,8 @@ contract DaoCreator {
         }
 
         // Add the DAO to the tracking registry
-        daoTracker.track(avatar, controller);
+        require(_daoTracker != address(0));
+        _daoTracker.track(avatar, controller);
 
         // Transfer ownership:
         avatar.transferOwnership(address(controller));
--- a/contracts/utils/DAOTracker.sol
+++ b/contracts/utils/DAOTracker.sol
@@ -1,0 +1,86 @@
+pragma solidity ^0.5.4;
+
+import "openzeppelin-solidity/contracts/ownership/Ownable.sol";
+
+/**
+ * @title An on-chain "source of truth" for what DAOs
+ *        should be index into DAOstack's subgraph.
+ */
+contract DAOTracker is Ownable {
+
+  struct TrackingInfo {
+    DAOToken            nativeToken;
+    Reputation          nativeReputation;
+    ControllerInterface controller;
+
+    // `blacklist` this from the subgraph's cache
+    bool blacklist;
+  }
+
+  // Mapping from Avatar address to a flag letting the subgraph
+  // know if it should track this DAO or not.
+  mapping(address=>TrackingInfo) public trackingInfo;
+
+  event TrackDAO(address indexed _avatar, address indexed _controller);
+
+  modifier onlyAvatarOwner(Avatar avatar) {
+    require(avatar.owner() == msg.sender,
+            "The caller must be the owner of the Avatar.");
+    _;
+  }
+
+  /**
+  * @dev track a new organization. This function will tell the subgraph
+  *      to start ingesting events from the DAO's contracts.
+  *      NOTE: This function should be called as early as possible in the deployment
+  *      process. Smart Contract Events that are emitted from blocks prior to this function's
+  *      event being emitted WILL NOT be ingested into the subgraph, leading to an incorrect
+  *      cache. If this happens to you, please contact the subgraph maintainer (DAOstack).
+  *      They will need to manually add your DAO to the subgraph, and rebuild the cache.
+  * @param _avatar the organization avatar
+  * @param _controller the organization controller
+  */
+  function track(Avatar _avatar, ControllerInterface _controller)
+  public
+  onlyAvatarOwner(_avatar) {
+    // Only allow the information to be set once. In the case of a controller upgrades,
+    // the subgraph will be updated via the UpgradeController event.
+    require(_avatar != address(0));
+    require(_controller != address(0));
+    require(tracking[_avatar].nativeToken == address(0));
+    require(tracking[_avatar].nativeReputation == address(0));
+    require(tracking[_avatar].controller == address(0));
+    require(tracking[_avatar].blacklist == false);
+
+    tracking[_avatar].nativeToken = _avatar.nativeToken();
+    tracking[_avatar].nativeReputation = _avatar.nativeReputation();
+    tracking[_avatar].controller = _controller;
+
+    emit TrackDAO(_avatar, _controller);
+  }
+
+  /**
+  * @dev blacklist a DAO from the cache. This should be callable by maintainer of the cache.
+  *      Blacklisting can be used to defend against DoS attacks, or to remove spam. In order
+  *      for this blacklisting to take affect within the cache, it would need to be rebuilt.
+  * @param _avatar the organization avatar
+  */
+  function blacklist(Avatar _avatar)
+  public
+  onlyOwner {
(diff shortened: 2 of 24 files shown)


ben-kaufman:
It looks like a good solution to me, and the code seems good as well, but since Oren is on vacation this week I prefer to wait for him with it since it's a relatively big change.

contracts/utils/DAOTracker.sol
```diff
@@ -0,0 +1,92 @@
...
+    onlyOwner {
+        require(_avatar != Avatar(0));
+        tracking[address(_avatar)].blacklist = true;
+    }
+
+    /**
+    * @dev reset a DAO in the cache. This should be callable by the maintainer of the cache.
+    *      The use case for this is extraordinary circumstances.
+    */
+    function reset(Avatar _avatar)
```
orenyodfat: who should be the owner of the tracker ? another dao ?

contracts/utils/DAOTracker.sol
```diff
@@ -0,0 +1,92 @@
...
+
+    /**
+    * @dev blacklist a DAO from the cache. This should be callable by maintainer of the cache.
+    *      Blacklisting can be used to defend against DoS attacks, or to remove spam. In order
+    *      for this blacklisting to take affect within the cache, it would need to be rebuilt.
+    * @param _avatar the organization avatar
+    */
+    function blacklist(Avatar _avatar)
+    public
+    onlyOwner {
```
orenyodfat: who should be the owner of the tracker ? another dao ?

contracts/utils/DAOTracker.sol
```diff
@@ -0,0 +1,92 @@
...
+    *      process. **Smart Contract Events that are emitted from blocks prior to this function's
+    *      event being emitted WILL NOT be ingested into the subgraph**, leading to an incorrect
+    *      cache. If this happens to you, please contact the subgraph maintainer. Your DAO will
+    *      need to be added to the subgraph's startup config, and the cache will need to be rebuilt.
+    * @param _avatar the organization avatar
+    * @param _controller the organization controller
+    */
+    function track(Avatar _avatar, ControllerInterface _controller)
+    public
+    onlyAvatarOwner(_avatar) {
```
orenyodfat: so if there is not call for track before transferring ownership to the controller , the dao cannot be tracked ?

orenyodfat:
how this is different from the daoregistery solution at https://github.com/daostack/arc-hive ?

contracts/utils/DAOTracker.sol
dOrgJelli: Currently since DAOstack is operating the caching server, they would be the owner of this contract. In the future this could be a collective effort.

contracts/utils/DAOTracker.sol
dOrgJelli: Currently since DAOstack is operating the caching server, they would be the owner of this contract. In the future this could be a collective effort.

contracts/utils/DAOTracker.sol
dOrgJelli: With this logic yes. I wanted to enforce this to ensure no other setup has taken place within the controller (adding schemes for ex).

dOrgJelli:
> how this is different from the daoregistery solution at https://github.com/daostack/arc-hive ?

Here's something you had said in our discussion that's linked to above: "The daoregistery is used by alchemy to fillet daos ,you can register dao on any time... not sure that tie that to indexing the dao is best option.".

I definitely agree with this, I don't think coupling a DAO registry with indexing makes sense.

dOrgJelli:
After talking with @orenyodfat (who spoke with @jellegerbrandy), it looks like a better / safer solution for indexing new DAOs is standing up a new DAO specific subgraph for them. I wrote something about how this might be done using Apollo Federations here: https://github.com/daostack/subgraph/issues/270

Please let me know what your thoughts are and feel free to close this PR if you've decided to not take the "DAOTracker approach".

dOrgJelli:
@orenyodfat @jellegerbrandy @leviadam 

Anything I can do to help get this merged? Once it's merged, I can work on adding support in the migration scripts, and then the subgraph. Then newly deployed DAOs will be automatically in Alchemy :)

contracts/utils/DAOTracker.sol
```diff
@@ -0,0 +1,92 @@
...
+    struct TrackingInfo {
+        DAOToken            nativeToken;
+        Reputation          nativeReputation;
+        ControllerInterface controller;
+
+        // `blacklist` this from the subgraph's cache
+        bool blacklist;
+    }
+
+    // Mapping from Avatar address to a flag letting the subgraph
```
orenyodfat: Mapping from Avatar address to TrackingInfo

contracts/utils/DAOTracker.sol
```diff
@@ -0,0 +1,92 @@
...
+ *        should be index into DAOstack's subgraph.
+ */
+contract DAOTracker is Ownable {
+
+    struct TrackingInfo {
+        DAOToken            nativeToken;
+        Reputation          nativeReputation;
+        ControllerInterface controller;
+
+        // `blacklist` this from the subgraph's cache
```
orenyodfat: `blacklist` this from the subgraph's indexing ?
this comment is not so clear

contracts/utils/DAOTracker.sol
```diff
@@ -0,0 +1,92 @@
...
+    *      cache. If this happens to you, please contact the subgraph maintainer. Your DAO will
+    *      need to be added to the subgraph's startup config, and the cache will need to be rebuilt.
+    * @param _avatar the organization avatar
+    * @param _controller the organization controller
+    */
+    function track(Avatar _avatar, ControllerInterface _controller)
+    public
+    onlyAvatarOwner(_avatar) {
+        // Only allow the information to be set once. In the case of a controller upgrades,
+        // the subgraph will be updated via the UpgradeController event.
```
orenyodfat: should we add a daoTrack event upon UpgradeController event ?

contracts/utils/DAOTracker.sol
```diff
@@ -0,0 +1,92 @@
...
+        require(_avatar != Avatar(0));
+        require(_controller != ControllerInterface(0));
+        require(tracking[address(_avatar)].nativeToken == DAOToken(0));
+        require(tracking[address(_avatar)].nativeReputation == Reputation(0));
+        require(tracking[address(_avatar)].controller == ControllerInterface(0));
```
orenyodfat: is it enough to require only require(tracking[address(_avatar)].controller == ControllerInterface(0)) ?

the other might be redundant.

contracts/utils/DAOTracker.sol
```diff
@@ -0,0 +1,92 @@
...
+        emit TrackDAO(address(_avatar), address(_controller));
+    }
+
+    /**
+    * @dev blacklist a DAO from the cache. This should be callable by maintainer of the cache.
+    *      Blacklisting can be used to defend against DoS attacks, or to remove spam. In order
+    *      for this blacklisting to take affect within the cache, it would need to be rebuilt.
+    * @param _avatar the organization avatar
+    */
+    function blacklist(Avatar _avatar)
```
orenyodfat: do we need an event here. ?

contracts/utils/DAOTracker.sol
```diff
@@ -0,0 +1,92 @@
...
+
+        // `blacklist` this from the subgraph's cache
+        bool blacklist;
+    }
+
+    // Mapping from Avatar address to a flag letting the subgraph
+    // know if it should track this DAO or not.
+    mapping(address=>TrackingInfo) public tracking;
+
+    event TrackDAO(address indexed _avatar, address indexed _controller);
```
orenyodfat: how about adding reputation and daotoken to the event ?
this will save us a call to the bc in the subgraph.

contracts/utils/DAOTracker.sol
```diff
@@ -0,0 +1,92 @@
...
+        Reputation          nativeReputation;
+        ControllerInterface controller;
+
+        // `blacklist` this from the subgraph's cache
+        bool blacklist;
+    }
+
+    // Mapping from Avatar address to a flag letting the subgraph
+    // know if it should track this DAO or not.
+    mapping(address=>TrackingInfo) public tracking;
```
orenyodfat: does this storage is usage is a must ? 
the state can be track via the event

contracts/utils/DAOTracker.sol
dOrgJelli: Yes you're right, these are redundant :P

contracts/utils/DAOTracker.sol
dOrgJelli: Added

contracts/utils/DAOTracker.sol
dOrgJelli: Added

contracts/utils/DAOTracker.sol
dOrgJelli: Not sure what you mean by this?

contracts/utils/DAOTracker.sol
dOrgJelli: Fixed

contracts/utils/DAOTracker.sol
dOrgJelli: Fixed

contracts/utils/DAOTracker.sol
dOrgJelli: Yes you're right. The situation where we'd want something on-chain is if we wanted to check the live state during a rebuild of the database. For example:
1. BadDAO adds itself
2. DAOstack blacklists BadDAO
3. The live subgraph receives the `BlacklistDAO` event and does its best to clear out all traces of BadDAO data.
4. BadDAO exploited a bug in the subgraph, so we need to rebuild the cache. When we rebuild and process BadDAO's `TrackDAO` event, we check the live state of the chain to see if it has been blacklisted, and if so we skip it.
```typescript
function handleTrackDAO(event: TrackDAO) {
  ...
  const info = daoTrackerSC.tracking(event._avatar).call({ block: "latest" })

  // Don't let BadDAO do what it did last time...
  if (info.blacklist) {
    return;
  }
}
```

dOrgJelli: I've minimized the storage that's in use to just a `mapping(address=>boolean)` to support the above use case.

contracts/utils/DAOTracker.sol
orenyodfat: it will be easier for the subgraph to handle one event (TraclDAP) .so by emitting TrackDAO event from the UpgradeController function the subgraph will be happier.

contracts/utils/DAOTracker.sol
```diff
@@ -0,0 +1,90 @@
+pragma solidity ^0.5.8;
```
orenyodfat: pragma solidity ^0.5.11;

contracts/universalSchemes/DaoCreator.sol
```diff
@@ -198,16 +200,19 @@ contract DaoCreator {
...
         } else {
             controller = _uController;
-            avatar.transferOwnership(address(controller));
-            // Transfer ownership:
-            nativeToken.transferOwnership(address(controller));
-            nativeReputation.transferOwnership(address(controller));
+        }
+
+        // Add the DAO to the tracking registry
+        daoTracker.track(avatar, controller);
```
orenyodfat: how about moving that after transfer ownership to the controller and pass to daoTracker.track only the avatar ?

contracts/utils/DAOTracker.sol
```diff
@@ -0,0 +1,90 @@
...
+    *      to start ingesting events from the DAO's contracts.
+    *      NOTE: This function should be called as early as possible in the DAO deployment
+    *      process. **Smart Contract Events that are emitted from blocks prior to this function's
+    *      event being emitted WILL NOT be ingested into the subgraph**, leading to an incorrect
+    *      cache. If this happens to you, please contact the subgraph maintainer. Your DAO will
+    *      need to be added to the subgraph's startup config, and the cache will need to be rebuilt.
+    * @param _avatar the organization avatar
+    * @param _controller the organization controller
+    */
+    function track(Avatar _avatar, ControllerInterface _controller)
```
orenyodfat: it seems to be safer and cleaner to pass in only avatar and get the controller as the avatar owner.

dOrgJelli: The reason for not doing this is because I wanted to make sure the `track(...)` function was being called by the current owner of the Avatar, and not some random person / caller. See modifier `onlyAvatarOwner`. I saw this as a good safety precaution that would ensure DAOs that don't want to be indexed, couldn't be.

contracts/utils/DAOTracker.sol
dOrgJelli: fixed

contracts/universalSchemes/DaoCreator.sol
dOrgJelli: See response for question about removing the `_controller` parameter. The conclusion there dictates whether this change is possible.

contracts/utils/DAOTracker.sol
dOrgJelli: I view these as two completely separate cases requiring two sets of independent logic. I think it's valuable to keep them separate so the subgraph mappings can choose what it wants to do.

dOrgJelli: Please let me know if this isn't the case ^^^

contracts/utils/DAOTracker.sol
orenyodfat: though in this case a dao which is created by daoCreator.sol will be index .want it or not .
dao which would like not to be index ,(not sure why he wants that), will be needed to be deployed manually and loose the advantage of using daoCreator.sol .

if we are convinced ,which I am still not, that there is a dao which cares if it is indexed or not , 
we can support that by passing a parameter to the daoCreator.sol forge func - bool _index ?

dOrgJelli: I suppose I misspoke when I said "...a good safety precaution that would ensure DAOs that don't want to be indexed, couldn't be".

What I really meant to say was "I saw this as a good safety precaution that would ensure the DAO being tracked was in the process of being constructed, and wasn't already live".

I see both as benefits to ensuring msg.sender is avatar.owner(), would you agree?

orenyodfat: make sense 👍

PR #644, opened 2019-07-11
Redeemer : call redeemDaoBounty only if pstate is Executed.

### LINKED ISSUE #643: Redeemer.redeem fails in proposals that are expiredinQueue
If the propsal state is `expiredInQueue`, the redeemer contracts `redeem()`  [will call genesisProtocol.redeemDaoBounty](https://github.com/daostack/arc/blob/master/contracts/utils/Redeemer.sol#L70).

However, `genesisProtocol.redeemDaoBounty` requires that the proposal be in the `Executed` state: https://github.com/daostack/infra/blob/master/contracts/votingMachines/GenesisProtocolLogic.sol#L401

And so the transaction will fail.

orenyodfat:
fix #643

### DIFF
--- a/test/contributionreward.js
+++ b/test/contributionreward.js
@@ -623,10 +623,11 @@ contract('ContributionReward', accounts => {
        await testSetup.contributionRewardParams.votingMachine.genesisProtocol.vote(proposalId,1,0,helpers.NULL_ADDRESS,{from:accounts[1]});
        await helpers.increaseTime(60+1);
        var arcUtils = await Redeemer.new(testSetup.contributionReward.address,testSetup.contributionRewardParams.votingMachine.genesisProtocol.address);
-       var proposal = await testSetup.contributionRewardParams.votingMachine.genesisProtocol.proposals(proposalId);
-       assert.equal(proposal.state,3); //ExpiredInQueue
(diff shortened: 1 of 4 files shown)


test/contributionreward.js
```diff
@@ -601,6 +601,35 @@ contract('ContributionReward', accounts => {
...
+                                                                      testSetup.standardTokenMock.address,
+                                                                      otherAvatar.address
+                                                                    );
+       var proposalId = await helpers.getValueFromLogs(tx, '_proposalId',1);
+
+       await testSetup.contributionRewardParams.votingMachine.genesisProtocol.vote(proposalId,1,0,helpers.NULL_ADDRESS,{from:accounts[1]});
+       await helpers.increaseTime(60+1);
+       var arcUtils = await Redeemer.new(testSetup.contributionReward.address,testSetup.contributionRewardParams.votingMachine.genesisProtocol.address);
+       var proposal = await testSetup.contributionRewardParams.votingMachine.genesisProtocol.proposals(proposalId);
+       assert.equal(proposal.state,3); //ExpiredInQueue
```
jellegerbrandy: state 3 is `Queued`, i think?

PR #646, opened 2019-07-15
Avatar: genericCall require call to contract

orenyodfat:


leviadam:
I'm not sure about this.
As part of our thoughts to use mostly generic call, the call should also work for non-contracts.

PR #652, opened 2019-08-02
Use Merkle Trees for RepAllocation addBeneficiaries method

k06a:
You can store only Merkle Root (32 bytes) inside the contract and make receivers to provide Merkle Proof as well as their address and amount information. Original addresses and amounts array can also be sent into events/logs to allow easily build Merkle Tree and proofs for everyone.

### DIFF
--- a/contracts/utils/RepAllocation.sol
+++ b/contracts/utils/RepAllocation.sol
@@ -14,6 +14,7 @@ contract RepAllocation is Ownable {
 
        // beneficiary -> amount
     mapping(address   =>   uint256) public reputationAllocations;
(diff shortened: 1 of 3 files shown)


orenyodfat:
does this the same as #653 ?

k06a:
@orenyodfat it is the same but without pushing all merkle tree leafs to the events. End require you to store all this leafs ofchain.

orenyodfat:
> @orenyodfat it is the same but without pushing all merkle tree leafs to the events. End require you to store all this leafs offchain.

so I guess this one can be closed as #653 include it.

PR #653, opened 2019-08-02
Use Merkle Trees for RepAllocation addBeneficiaries method (+EVENTS)

k06a:


leviadam:
Thanks @k06a !
I like this PR very much. Still did not have time to review it carefully. But it is a very good functionality.

@orenyodfat @ben-kaufman ?

contracts/utils/RepAllocation.sol
```diff
@@ -42,6 +44,48 @@ contract RepAllocation is Ownable {
...
+    }
+
+    /**
+     * @dev add addBeneficiariesRoot function
+     * https://github.com/OpenZeppelin/openzeppelin-contracts/blob/master/contracts/cryptography/MerkleProof.sol
+     * @param _beneficiary user address
+     * @param _amount allocation
+     * @param _proof Merkle proof related to previously submitted Merkle root
+     */
+    function revealBeneficiary(address _beneficiary, uint256 _amount, bytes32[] memory _proof) public {
```
orenyodfat: how about using openzeppline MarkleProof lib verify function ? 
https://github.com/OpenZeppelin/openzeppelin-contracts/blob/master/contracts/cryptography/MerkleProof.sol#L13

contracts/utils/RepAllocation.sol
```diff
@@ -42,6 +44,48 @@ contract RepAllocation is Ownable {
...
+        address[] memory _beneficiaries,
+        uint256[] memory _amounts
+    ) public onlyOwner {
+        require(!reputationAllocationsRoots[_root]);
+        reputationAllocationsRoots[_root] = true;
+        emit BeneficiaryAddressAddedByRoot(_root, _beneficiaries, _amounts);
+    }
+
+    /**
+     * @dev add addBeneficiariesRoot function
```
orenyodfat: revealBeneficiary (not addBeneficiariesRoot)

contracts/utils/RepAllocation.sol
k06a: It can be used, I used my own because think it will be useful to recover `index` for invalidation. Now I see invalidation is done inside `addBeneficiary`.

k06a: But you should also add root argument, because OZ implementation only verifies root matches, do not return recovered root.

k06a: Fixed

PR #658, opened 2019-09-08
Add NectarRepAllocation.sol

orenyodfat:
par of bootstrap scheme for nectar Dao

### DIFF
--- a/contracts/test/NectarRepAllocation.sol
+++ b/contracts/test/NectarRepAllocation.sol
@@ -1,0 +1,66 @@
+pragma solidity ^0.5.11;
+
+import "openzeppelin-solidity/contracts/math/SafeMath.sol";
+
+contract MiniMeToken {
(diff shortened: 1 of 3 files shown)


contracts/test/NectarRepAllocation.sol
```diff
@@ -0,0 +1,66 @@
...
+        external
+    {
+        require(token == MiniMeToken(0), "can be called only one time");
+        require(_token != MiniMeToken(0), "token cannot be zero");
+        token = _token;
+        reputationReward = _reputationReward;
+        claimingStartTime = _claimingStartTime;
+        claimingEndTime = _claimingEndTime;
+        blockReference = _blockReference;
+        if ((claimingStartTime != 0) || (claimingEndTime != 0)) {
```
leviadam: This means I cannot have a start with no end, right?

orenyodfat: right. 
though you can have end time without start time.

PR #659, opened 2019-09-08
Add non universal genericScheme

orenyodfat:


dkent600:
Just wondering:  Why is this needed (what are its advantages)?

orenyodfat:
- Less complex
- More secure
- Allow easy support to register multiple schemes instances with the same
name to the same dao .(by attaching specific aliases to each ).

PR #666, opened 2019-09-19
emit redeem event per batch + comments/naming

### LINKED ISSUE #665: Redeem event in ContinuousLocking4Reputation
https://github.com/daostack/arc/blob/a409de0a597c551bda39975951753d3d2a529131/contracts/schemes/ContinuousLockingToken4Reputation.sol#L21

Should we emit a redeem event for each batch which is redeemed to make it easy on the UI?

orenyodfat:
easy to get redeem per patch for the ui layer.
fix #665 

address #667


contracts/schemes/ContinuousLockingToken4Reputation.sol
```diff
@@ -67,16 +67,19 @@ contract ContinuousLocking4Reputation is Agreement {
      * @param _avatar the avatar to mint reputation from
      * @param _reputationReward the reputation reward per locking batch that this contract will reward
```
jellegerbrandy: This value is assinged to _reputationRewardLeft, which is a global  variable, so (it seems to me) this is not the reputatino reward per locking batch, butthe TOTAL amount of reputation that can be minted by this contract.

PR #687, opened 2019-11-14
Offchain concat

ben-kaufman:


test/helpers.js
```diff
@@ -345,3 +345,11 @@ export const increaseTime = async function(duration) {
     });
   });
 };
+
+export const concatBytes = function (bytes1, bytes2) {
+  return bytes1 + (bytes2.slice(2));
```
orenyodfat: Does this assume that the bytes does not include the bytes header (which include the bytes length) as in solidity type bytes ?

ben-kaufman: Not sure I understand the question... This simply assumes the bytes format which starts like this: `"0x..."`.

orenyodfat:
commit 1e6a6e4 update a test which cause it to fail. @ben-kaufman

PR #688, opened 2019-11-18
daofactory : support forge a dao from specific version

### LINKED ISSUE #686: arc-experimental : daofactory to support dao instance from older version.
https://forum.openzeppelin.com/t/create-a-new-instance-from-older-package-version/1747

orenyodfat:
fix #686

### DIFF
--- a/contracts/utils/DAOFactory.sol
+++ b/contracts/utils/DAOFactory.sol
@@ -16,11 +16,23 @@ contract DAOFactory is Initializable {
     event NewOrg (address indexed _avatar);
     event InitialSchemesSet (address indexed _avatar);
     event SchemeInstance(address indexed _scheme, string indexed _name);
+    /**
+    * @dev Emitted when a new proxy is created.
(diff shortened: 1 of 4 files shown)


contracts/utils/DAOFactory.sol
```diff
@@ -16,11 +16,23 @@ contract DAOFactory is Initializable {
...
+        address sender;
+        uint64[3] packageVersion;
+    }
 
-    mapping(address=>address) public locks;
+    mapping(address=>Locks) public locks;
     App public app;
     string public constant PACKAGE_NAME = "DAOstack";
     DAOTracker private daoTracker;
+    uint64[3] private packageVersion;
```
ben-kaufman: Why is that a storage variable. Wouldn't it be cheaper (and better) to just pass it to the `_forgeOrg` function?

orenyodfat: it will be cheaper , though this is here due to  "stack is too deep" issue .

PR #697, opened 2019-12-03
Competition contract + ContributionRewardExt scheme

orenyodfat:


contracts/schemes/Competition.sol
```diff
@@ -0,0 +1,349 @@
...
+    * @return proposalId the proposal id.
+    */
+    function proposeCompetition(
+            string calldata _descriptionHash,
+            int256 _reputationChange,
+            uint[3] calldata _rewards,
+            IERC20 _externalToken,
+            uint256[] calldata _rewardSplit,
+            uint256[4] calldata _competitionParams,
+            address payable _contributionRewardExt //address of the contract to redeem from.
```
ben-kaufman: This is an attack vector with multiple possible exploits. This parameter should probably be part of the `initialize()` method.

contracts/schemes/ContributionRewardExt.sol
```diff
@@ -0,0 +1,455 @@
...
+            require(msg.sender == rewarder, "only rewarder allowed to redeem");
+        }
+        _;
+    }
+
+    mapping(bytes32=>ContributionProposal) public organizationProposals;
+
+    IntVoteInterface public votingMachine;
+    bytes32 public voteParams;
+    address public contractToCall;
```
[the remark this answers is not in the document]
orenyodfat: removed

contracts/schemes/Competition.sol
```diff
@@ -0,0 +1,349 @@
...
+        /** get how many elements are greater than a given element*/
+        for (uint256 i; i < topSuggestionsOrdered.length; i++) {
+            if (suggestions[topSuggestionsOrdered[i]].totalVotes > suggestions[_suggestionId].totalVotes) {
+                index++;
+            }
+        }
+    }
+
+    /**
+    * @dev refreshTopSuggestions this function maintain a winners list array.
```
jellegerbrandy: Perhaps be more specific, and write, refreshTopSuggestions will check if the given suggestion is among the top suggestions, and if so, update the  list of top suggestions

contracts/schemes/Competition.sol
```diff
@@ -0,0 +1,349 @@
...
+    * @param _proposalId proposal id
```
jellegerbrandy: this param is not needed, as we can derive it from the suggestion.

contracts/schemes/Competition.sol
```diff
@@ -0,0 +1,349 @@
...
+    }
+
+    /**
+    * @dev redeem a winning suggestion reward
+    * @param _suggestionId suggestionId
+    * @param _beneficiary - the reward beneficiary
+    */
+    function _redeem(uint256 _suggestionId, address payable _beneficiary) private {
+        bytes32 proposalId = suggestions[_suggestionId].proposalId;
+        require(proposalId != bytes32(0), "suggestion not exist");
```
jellegerbrandy: "suggestion does not exist"

contracts/schemes/Competition.sol
```diff
@@ -0,0 +1,349 @@
...
+                uint256 orderIndex = getOrderedIndexOfSuggestion(proposalId, _suggestionId);
+                suggestions[_suggestionId].suggester = address(0);
+                uint256 rewardPercentage = 0;
+                uint256 numberOfTieSuggestions = proposal.tiesSuggestions[suggestions[_suggestionId].totalVotes];
+                uint256 j;
+                //calc the reward percentage for this suggestion
+                for (j = orderIndex; j < (orderIndex+numberOfTieSuggestions); j++) {
+                    rewardPercentage = rewardPercentage.add(proposal.rewardSplit[j]);
+                }
+                rewardPercentage = rewardPercentage.div(numberOfTieSuggestions);
```
jellegerbrandy: you may get some rounding errors here and distribute too litte. (i.e. if the reward percentage is 5 numberofTies is 3, each will getonly 1 percent.

contracts/schemes/Competition.sol
```diff
@@ -0,0 +1,349 @@
...
+        require(proposalId != bytes32(0), "suggestion not exist");
+        Proposal storage proposal = proposals[proposalId];
+        // solhint-disable-next-line not-rely-on-time
+        require(proposal.endTime < now, "competition is still on");
+        uint256 amount;
+        //check if there is a win
+        for (uint256 i = 0; i < proposal.topSuggestionsOrdered.length; i++) {
+            if (suggestions[proposal.topSuggestionsOrdered[i]].suggester != address(0)) {
+                uint256 orderIndex = getOrderedIndexOfSuggestion(proposalId, _suggestionId);
+                suggestions[_suggestionId].suggester = address(0);
```
jellegerbrandy: I do not really get what this is supposed to do - is this to mark that the suggester has been paid?  If so, where is the check that she cannot be paid twice?

contracts/schemes/Competition.sol
```diff
@@ -0,0 +1,349 @@
...
+            if (suggestions[topSuggestionsOrdered[smallest]].totalVotes < suggestions[_suggestionId].totalVotes) {
+                topSuggestionsOrdered[smallest] = _suggestionId;
+            }
+        }
+    }
+
+    /**
+    * @dev redeem a winning suggestion reward
+    * @param _suggestionId suggestionId
+    * @param _beneficiary - the reward beneficiary
```
jellegerbrandy: honestly, the code in this fucntion is too complex for me to understand. But i do see it pays out to _beneficiary, but it does not seem to check any relation between the suggester of the suggestion and either msg.sender or _beneficiary.

contracts/schemes/Competition.sol
```diff
@@ -0,0 +1,349 @@
...
+                for (j = orderIndex; j < (orderIndex+numberOfTieSuggestions); j++) {
+                    rewardPercentage = rewardPercentage.add(proposal.rewardSplit[j]);
+                }
+                rewardPercentage = rewardPercentage.div(numberOfTieSuggestions);
+                uint256 rewardPercentageLeft = 0;
+                if (proposal.topSuggestionsOrdered.length < proposal.numberOfWinners) {
+                    for (j = proposal.topSuggestionsOrdered.length; j < proposal.numberOfWinners; j++) {
+                        rewardPercentageLeft = rewardPercentageLeft.add(proposal.rewardSplit[j]);
+                    }
+                    //if there are less winners than the proposal number of winners so divide the pre allocated
```
jellegerbrandy: move this comment 3 lines up..

contracts/schemes/Competition.sol
```diff
@@ -0,0 +1,349 @@
...
+    function _redeem(uint256 _suggestionId, address payable _beneficiary) private {
+        bytes32 proposalId = suggestions[_suggestionId].proposalId;
+        require(proposalId != bytes32(0), "suggestion not exist");
+        Proposal storage proposal = proposals[proposalId];
+        // solhint-disable-next-line not-rely-on-time
+        require(proposal.endTime < now, "competition is still on");
+        uint256 amount;
+        //check if there is a win
+        for (uint256 i = 0; i < proposal.topSuggestionsOrdered.length; i++) {
+            if (suggestions[proposal.topSuggestionsOrdered[i]].suggester != address(0)) {
```
jellegerbrandy: I do not see at all why this iteration is needed at all - can you add  a comment?

jellegerbrandy: (seems to me the for loop can be simply removed)

contracts/schemes/Competition.sol
```diff
@@ -0,0 +1,357 @@
...
+        setSnapshotBlock(proposalId);
+        Proposal storage proposal = proposals[proposalId];
+        // solhint-disable-next-line not-rely-on-time
+        require(proposal.endTime > now, "competition ended");
+        Suggestion storage suggestion = suggestions[_suggestionId];
+        require(suggestion.votes[msg.sender] == 0, "already voted on this suggestion");
+        require(proposal.votesPerVoter[msg.sender] < proposal.numberOfVotesPerVoters,
+        "exceed number of votes allowed");
+        proposal.votesPerVoter[msg.sender] = proposal.votesPerVoter[msg.sender].add(1);
+        suggestion.votes[msg.sender] = suggestion.votes[msg.sender].add(1);
```
ben-kaufman: What is this line for? It is always 1 and always rewritten in the function a few lines below.

contracts/schemes/Competition.sol
```diff
@@ -0,0 +1,349 @@
...
+        uint256 snapshotBlock;
+        //contributionRewardExt
+        uint256 reputationReward;
+        uint256 ethReward;
+        uint256 nativeTokenReward;
+        uint256 externalTokenReward;
+
+        uint256[] topSuggestionsOrdered;
+        //mapping from suggestions totalVotes to the number of suggestions with the same totalVotes.
+        mapping(uint256=>uint256) tiesSuggestions;
```
jellegerbrandy: Actually, this mapping will  lso store suggestions that have a unique number of votes - so not just ties. Perhaps rename to `votesToSuggestions` or `suggestionsPerVote`.

contracts/schemes/Competition.sol
```diff
@@ -0,0 +1,349 @@
...
+        require(proposal.endTime > now, "competition ended");
+        Suggestion storage suggestion = suggestions[_suggestionId];
+        require(suggestion.votes[msg.sender] == 0, "already voted on this suggestion");
+        require(proposal.votesPerVoter[msg.sender] < proposal.numberOfVotesPerVoters,
+        "exceed number of votes allowed");
+        proposal.votesPerVoter[msg.sender] = proposal.votesPerVoter[msg.sender].add(1);
+        suggestion.votes[msg.sender] = suggestion.votes[msg.sender].add(1);
+        Avatar avatar = ContributionRewardExt(proposal.contributionRewardExt).avatar();
+        uint256 reputation = avatar.nativeReputation().balanceOfAt(msg.sender, proposals[proposalId].snapshotBlock);
+        require(reputation > 0, "voter has no reputation");
```
jellegerbrandy: This is a very confusing error message if you do have reputation, but did not have it in the snapshot. Change to` voter had no reputation when ..` (??)

contracts/schemes/Competition.sol
orenyodfat: removed with ced6964

contracts/schemes/Competition.sol
```diff
@@ -0,0 +1,349 @@
...
+    struct Proposal {
+        uint256 numberOfWinners;
+        uint256[] rewardSplit;
+        uint256 startTime;
+        uint256 votingStartTime;
+        uint256 endTime;
+        uint256 numberOfVotesPerVoters;
+        address payable contributionRewardExt;
+        uint256 snapshotBlock;
+        //contributionRewardExt
```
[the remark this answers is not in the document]
orenyodfat: will remove it

contracts/schemes/Competition.sol
```diff
@@ -0,0 +1,349 @@
...
+            }
+        }
+        _redeem(_suggestionId, beneficiary);
+    }
+
+    /**
+    * @dev setSnapshotBlock set the block for the reputaion snapshot
+    * @param _proposalId the proposal id
+    */
+    function setSnapshotBlock(bytes32 _proposalId) public {
```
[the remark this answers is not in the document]
orenyodfat: yes. 
-it can be called by any one 
-it is called on the first vote
-the block number can be set. one time
-it cannot be set before the voting time due to https://github.com/daostack/arc/blob/new_cr/contracts/schemes/Competition.sol#L249

contracts/schemes/Competition.sol
```diff
@@ -0,0 +1,349 @@
...
+    );
+
+    event SnapshotBlock(
+        bytes32 indexed _proposalId,
+        uint256 _snapshotBlock
+    );
+
+    // A struct holding the data for a competition proposal
+    struct Proposal {
+        uint256 numberOfWinners;
```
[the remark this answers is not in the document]
orenyodfat: right. I used the same approach in the proposal param.
I think it is easy this way

contracts/schemes/Competition.sol
```diff
@@ -0,0 +1,349 @@
...
+        uint256[] rewardSplit;
+        uint256 startTime;
+        uint256 votingStartTime;
+        uint256 endTime;
+        uint256 numberOfVotesPerVoters;
+        address payable contributionRewardExt;
+        uint256 snapshotBlock;
+        //contributionRewardExt
+        uint256 reputationReward;
```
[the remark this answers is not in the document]
orenyodfat: yes. 
though once the reward is redeemed from contributionReward it is set to zero.

contracts/schemes/Competition.sol
```diff
@@ -0,0 +1,349 @@
...
+    *         _competitionParams[0] - competition startTime
+    *         _competitionParams[1] - _votingStartTime competition voting start time
+    *         _competitionParams[2] - _endTime competition end time
+    *         _competitionParams[3] - _numberOfVotesPerVoters on how many suggestions a voter can vote
+    * @param _contributionRewardExt the contributionRewardExt scheme which
+    *        manage and allocate the rewards for the competition.
+    * @return proposalId the proposal id.
+    */
+    function proposeCompetition(
+            string calldata _descriptionHash,
```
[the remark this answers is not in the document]
orenyodfat: the compiler asked for it.
see https://solidity.readthedocs.io/en/v0.5.3/types.html#data-location

contracts/schemes/Competition.sol
```diff
@@ -0,0 +1,349 @@
...
+            bytes32 _proposalId,
+            string calldata _descriptionHash
+    )
+    external
+    returns(uint256)
+    {
+      // solhint-disable-next-line not-rely-on-time
+        require(proposals[_proposalId].startTime <= now, "competition not started yet");
+        // solhint-disable-next-line not-rely-on-time
+        require(proposals[_proposalId].endTime > now, "competition ended");
```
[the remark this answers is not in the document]
orenyodfat: due to the fogginess of the specification I will add a parameter : suggestionsEndTime

contracts/schemes/Competition.sol
```diff
@@ -0,0 +1,349 @@
...
+    function vote(uint256 _suggestionId)
+    external
+    returns(bool)
+    {
+        bytes32 proposalId = suggestions[_suggestionId].proposalId;
+        require(proposalId != bytes32(0), "suggestion not exist");
+        setSnapshotBlock(proposalId);
+        Proposal storage proposal = proposals[proposalId];
+        // solhint-disable-next-line not-rely-on-time
+        require(proposal.endTime > now, "competition ended");
```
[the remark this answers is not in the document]
orenyodfat: this is checked at https://github.com/daostack/arc/blob/new_cr/contracts/schemes/Competition.sol#L249

contracts/schemes/Competition.sol
```diff
@@ -0,0 +1,349 @@
...
+        require(proposal.votesPerVoter[msg.sender] < proposal.numberOfVotesPerVoters,
+        "exceed number of votes allowed");
+        proposal.votesPerVoter[msg.sender] = proposal.votesPerVoter[msg.sender].add(1);
+        suggestion.votes[msg.sender] = suggestion.votes[msg.sender].add(1);
+        Avatar avatar = ContributionRewardExt(proposal.contributionRewardExt).avatar();
+        uint256 reputation = avatar.nativeReputation().balanceOfAt(msg.sender, proposals[proposalId].snapshotBlock);
+        require(reputation > 0, "voter has no reputation");
+        suggestion.totalVotes = suggestion.totalVotes.add(reputation);
+        suggestion.votes[msg.sender] = reputation;
+        proposal.tiesSuggestions[suggestion.totalVotes] = proposal.tiesSuggestions[suggestion.totalVotes].add(1);
```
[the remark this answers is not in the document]
orenyodfat: I do not understand

contracts/schemes/Competition.sol
```diff
@@ -0,0 +1,349 @@
...
+        refreshTopSuggestions(proposalId, _suggestionId);
+        emit NewVote(proposalId, _suggestionId, msg.sender, reputation);
+        return true;
+    }
+
+    /**
+    * @dev redeem a winning suggestion reward
+    * @param _suggestionId suggestionId
+    * @param _beneficiary - the reward beneficiary.
```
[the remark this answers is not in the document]
orenyodfat: well, this is added here to enable the suggester to reward other account ..
especially if the reward is reputation which is not transferable

contracts/schemes/Competition.sol
orenyodfat: yes. though it is private function and the proposalId already been extracted.

contracts/schemes/Competition.sol
orenyodfat: we need to see what to do with this dust .
- send it back to the dao ?
-give it to the winner?

fixing the rounding issue might add some unneeded complexity

contracts/schemes/Competition.sol
orenyodfat: https://github.com/daostack/arc/blob/new_cr/contracts/schemes/Competition.sol#L313

contracts/schemes/Competition.sol
orenyodfat: the beneficiary is the suggester .
(or otherwise if the suggester want it differently)

orenyodfat: will simplify this function

contracts/schemes/ContributionRewardExt.sol
```diff
@@ -0,0 +1,454 @@
...
+            emit RedeemEther(address(avatar), _proposalId, _proposal.beneficiary, amount);
+        }
+    }
+
+    /**
+    * @dev RedeemNativeToken reward for proposal
+    * @param _proposalId the ID of the voting in the voting machine
+    * @return amount the external token redeemed amount
+    */
+    function redeemExternalToken(bytes32 _proposalId) public returns(uint256 amount) {
```
dkent600: wondering if we should be anticipating here the ability to reward multiple external tokens?

contracts/schemes/ContributionRewardExt.sol
```diff
@@ -0,0 +1,454 @@
...
+    public
+    onlyRewarder
+    {
+        ContributionProposal storage proposal = organizationProposals[_proposalId];
+        require(proposal.executionTime != 0);
+        //this will ensure sum zero of reputation.
+        proposal.reputationChangeLeft = proposal.reputationChangeLeft.sub(_reputation);
+        require(
+        Controller(
+        avatar.owner()).mintReputation(_reputation, _beneficiary, address(avatar)));
```
dkent600: Should the mint be inside of `if (_reputation != 0)`?

Should we not burn when if (_reputation < 0)?

contracts/schemes/ContributionRewardExt.sol
orenyodfat: we might add that later on.not in this phase.

contracts/schemes/ContributionRewardExt.sol
orenyodfat: > Should the mint be inside of if (_reputation != 0)

we can but not must. I will add that.

> Should we not burn when if (_reputation < 0)?

the _reputation here is uint so it cannot be < 0

dkent600: See `redeemReputation`.  Why is `_reputation ` not int256 ?  (and perhaps name it without the underscore, to be consistent with `redeemReputation`)

contracts/schemes/ContributionRewardExt.sol
```diff
@@ -0,0 +1,454 @@
...
+    */
+    function redeemReputation(bytes32 _proposalId) public returns(int256 reputation) {
+        ContributionProposal memory _proposal = organizationProposals[_proposalId];
+        ContributionProposal storage proposal = organizationProposals[_proposalId];
+        require(proposal.executionTime != 0);
+        //set proposal reward to zero to prevent reentrancy attack.
+        proposal.reputationChange = 0;
+
+        if (_proposal.beneficiary == address(this)) {
+            if (_proposal.reputationChange > 0) {//for now only mint(not burn) rep allowed from ext contract.
```
dkent600: Why is burn not being allowed?

contracts/schemes/ContributionRewardExt.sol
dkent600: can we comment, like we did in `redeemReputation`, that burn is not currently allowed?

contracts/schemes/ContributionRewardExt.sol
```diff
@@ -0,0 +1,454 @@
...
+    * @param _beneficiary Who gets the rewards. if equal to 0 the beneficiary will be msg.sender.
+    * @param _proposer proposer . if equal to 0 the proposer will be msg.sender.
+    */
+    function proposeContributionReward(
+        string memory _descriptionHash,
+        int256 _reputationChange,
+        uint[3] memory _rewards,
+        IERC20 _externalToken,
+        address payable _beneficiary,
+        address _proposer
```
ben-kaufman: Why would you want to allow anyone to propose in the name of others?

contracts/schemes/ContributionRewardExt.sol
```diff
@@ -0,0 +1,454 @@
...
+    function redeemReputation(bytes32 _proposalId) public returns(int256 reputation) {
+        ContributionProposal memory _proposal = organizationProposals[_proposalId];
+        ContributionProposal storage proposal = organizationProposals[_proposalId];
+        require(proposal.executionTime != 0);
+        //set proposal reward to zero to prevent reentrancy attack.
+        proposal.reputationChange = 0;
+
+        if (_proposal.beneficiary == address(this)) {
+            if (_proposal.reputationChange > 0) {//for now only mint(not burn) rep allowed from ext contract.
+                proposal.reputationChangeLeft = uint256(_proposal.reputationChange);
```
ben-kaufman: Why not just add a `return;` here to save gas on the code below?

contracts/schemes/ContributionRewardExt.sol
```diff
@@ -0,0 +1,454 @@
...
+
+        if (_proposal.beneficiary == address(this)) {
+            if (_proposal.externalTokenReward != 0) {
+                proposal.externalTokenRewardLeft = _proposal.externalTokenReward;
+            }
+        }
+
+        if (proposal.externalToken != IERC20(0) && _proposal.externalTokenReward > 0) {
+            amount = _proposal.externalTokenReward;
+            if (amount > 0) {
```
ben-kaufman: This `if` is duplicated, as it checks the same think as the `if` above, please remove.

contracts/schemes/ContributionRewardExt.sol
orenyodfat: because it is complicated the rewarder contract.and there is no real req re that yet

contracts/schemes/ContributionRewardExt.sol
orenyodfat: 1. because we need that to the competition .
2. to allow propose on behalf of someone else...e.g to get rep from hc ..to this address

contracts/schemes/ContributionRewardExt.sol
orenyodfat: redundant

contracts/schemes/Competition.sol
orenyodfat: added sendLeftOverFubnds at 
456cbff

contracts/schemes/ContributionRewardExt.sol
```diff
@@ -0,0 +1,452 @@
...
+
+import "@daostack/infra/contracts/votingMachines/IntVoteInterface.sol";
+import "@daostack/infra/contracts/votingMachines/VotingMachineCallbacksInterface.sol";
+import "../votingMachines/VotingMachineCallbacks.sol";
+import "../libs/SafeERC20.sol";
+
+
+/**
+ * @title A scheme for proposing and rewarding contributions to an organization
+ * @dev An agent can ask an organization to recognize a contribution and reward
```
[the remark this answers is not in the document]
jellegerbrandy: So (after reading the code) this is an extended version of the ContriutionReward contract, with the possibility to assign a rewarder, which, after the contributionreward has been accepted, can then later distribute the assets as it would like.

So perhaps call it "ContributionRewardWithRewarder"?

To me, it seems a bit crazy do stuff all of this logic in a single contract (it is an enormous, complex, contract) 

If I understand it correctly, if the rewards were all transferable tokens (i.e. no rep), we could have a much cleaner setup by using our original contribution reward, and transfer the assets to a beneficiary contract  that manages the payouts for the competition.  But (I think) this will not be possible bc the rights to mint rep are difficult to manage.

(which makes me think - not for this PR, but for arc2.0 - instead of doing all this administration of keeping track fo which schemes have the right to mint which rep, it could perhaps be easier and more secure to work with a "rightToMintRep"  token)

contracts/schemes/Competition.sol
```diff
@@ -0,0 +1,396 @@
...
+                if (suggestions[topSuggestions[i]].totalVotes <
+                    suggestions[topSuggestions[smallest]].totalVotes) {
+                    smallest = i;
+                } else if (topSuggestions[i] == _suggestionId) {
+                  //the suggestion is already in the topSuggestions list
+                    return;
+                }
+            }
+
+            if (suggestions[topSuggestions[smallest]].totalVotes < suggestions[_suggestionId].totalVotes) {
```
jellegerbrandy: So, if the _suggestionId has the **same** number of votes as the smallest in the topSuggestions, it will **not** be added to the topSuggestions. 

That seems unfair :-)

contracts/schemes/ContributionRewardExt.sol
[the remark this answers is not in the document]
orenyodfat: > So (after reading the code) this is an extended version of the ContriutionReward contract, with the possibility to assign a rewarder, which, after the contributionreward has been accepted, can then later distribute the assets as it would like.
> 
> So perhaps call it "ContributionRewardWithRewarder"?
> 
> To me, it seems a bit crazy do stuff all of this logic in a single contract (it is an enormous, complex, contract)
> 
> If I understand it correctly, if the rewards were all transferable tokens (i.e. no rep), we could have a much cleaner setup by using our original contribution reward, and transfer the assets to a beneficiary contract that manages the payouts for the competition. But (I think) this will not be possible bc the rights to mint rep are difficult to manage.
> 
> (which makes me think - not for this PR, but for arc2.0 - instead of doing all this administration of keeping track fo which schemes have the right to mint which rep, it could perhaps be easier and more secure to work with a "rightToMintRep" token)

yes. If a rep was not a req here we could teoreticlyl use a normal cr with the beneficiary the competition scheme / or other

contracts/schemes/ContributionRewardExt.sol
```diff
@@ -0,0 +1,452 @@
...
+
+    /**
+     * @dev initialize
+     * @param _avatar the avatar to mint reputation from
+     * @param _votingMachine the voting machines address to
+     * @param _voteParams voting machine parameters.
+     * @param _rewarder an address which allowed to redeem the contribution.
+       if _rewarder is 0 this param is agnored.
+     */
+    function initialize(
```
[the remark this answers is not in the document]
orenyodfat: eventually .yes. as for all none universals schemes

contracts/schemes/ContributionRewardExt.sol
```diff
@@ -0,0 +1,452 @@
...
+    onlyVotingMachine(_proposalId)
+    returns(bool) {
+        require(organizationProposals[_proposalId].executionTime == 0);
+        require(organizationProposals[_proposalId].beneficiary != address(0));
+        // Check if vote was successful:
+        if (_decision == 1) {
+          // solhint-disable-next-line not-rely-on-time
+            organizationProposals[_proposalId].executionTime = now;
+        }
+        emit ProposalExecuted(address(avatar), _proposalId, _decision);
```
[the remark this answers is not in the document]
orenyodfat: same as in legacy ContributionReward.

contracts/schemes/ContributionRewardExt.sol
```diff
@@ -0,0 +1,452 @@
...
+    * @dev execution of proposals, can only be called by the voting machine in which the vote is held.
+    * @param _proposalId the ID of the voting in the voting machine
+    * @param _decision a parameter of the voting result, 1 yes and 2 is no.
+    */
+    function executeProposal(bytes32 _proposalId, int256 _decision)
+    external
+    onlyVotingMachine(_proposalId)
+    returns(bool) {
+        require(organizationProposals[_proposalId].executionTime == 0);
+        require(organizationProposals[_proposalId].beneficiary != address(0));
```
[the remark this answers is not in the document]
orenyodfat: mmm. this was taken from  the universal `ContributionReward executeProposal ` .
It might be here to prevent execution by mistake(due to a bug or bad actor)  of a votingMachine of uninitilize proposal.
we need to give it another thought before removing that here.

contracts/schemes/ContributionRewardExt.sol
```diff
@@ -0,0 +1,452 @@
...
+
+        ContributionProposal memory proposal = ContributionProposal({
+            nativeTokenReward: _rewards[0],
+            reputationChange: _reputationChange,
+            ethReward: _rewards[1],
+            externalToken: _externalToken,
+            externalTokenReward: _rewards[2],
+            beneficiary: beneficiary,
+            executionTime: 0,
+            nativeTokenRewardLeft: 0,
```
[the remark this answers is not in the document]
orenyodfat: because we want to enable call to`redeemNativeTokenFromExtContract` only if the reward was redeemed .
- nativeTokenRewardLeft is set upon redeem.

contracts/schemes/ContributionRewardExt.sol
```diff
@@ -0,0 +1,452 @@
...
+
+    /**
+    * @dev RedeemReputation reward for proposal
+    * @param _proposalId the ID of the voting in the voting machine
+    * @return reputation the redeemed reputation.
+    */
+    function redeemReputation(bytes32 _proposalId) public returns(int256 reputation) {
+        ContributionProposal memory _proposal = organizationProposals[_proposalId];
+        ContributionProposal storage proposal = organizationProposals[_proposalId];
+        require(proposal.executionTime != 0);
```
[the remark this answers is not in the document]
orenyodfat: yes. now that we are not using period payments we do not need excutionTime

contracts/schemes/ContributionRewardExt.sol
```diff
@@ -0,0 +1,452 @@
...
+        //set proposal reward to zero to prevent reentrancy attack.
+        proposal.reputationChange = 0;
+
+        if (_proposal.beneficiary == address(this)) {
```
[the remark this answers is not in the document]
orenyodfat: `if the beneficiary is the current contract, we are not minting the rep to it but instead refer to a mechanism in which the rep can be minted by the current contract per request of the rewarder`

contracts/schemes/ContributionRewardExt.sol
```diff
@@ -0,0 +1,452 @@
...
+
+        ContributionProposal memory _proposal = organizationProposals[_proposalId];
+        ContributionProposal storage proposal = organizationProposals[_proposalId];
+        require(proposal.executionTime != 0);
+        //set proposal rewards to zero to prevent reentrancy attack.
+        proposal.nativeTokenReward = 0;
+
+        if (_proposal.beneficiary == address(this)) {
+            if (_proposal.nativeTokenReward != 0) {
+                proposal.nativeTokenRewardLeft = _proposal.nativeTokenReward;
```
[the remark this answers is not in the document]
orenyodfat: because you want it to be set only one time..upon redeem.

contracts/schemes/ContributionRewardExt.sol
```diff
@@ -0,0 +1,452 @@
...
+                address(proposal.externalToken).safeTransfer(_beneficiary, _amount);
+                emit RedeemExternalToken(address(avatar), _proposalId, _beneficiary, _amount);
+            }
+        }
+    }
+
+    /**
+    * @dev redeem rewards for proposal
+    * @param _proposalId the ID of the voting in the voting machine
+    * @param _whatToRedeem whatToRedeem array:
```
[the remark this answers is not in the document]
orenyodfat: this is an helper function to enable redeem of all in one tx.
the `fine control` is needed to avoid reverts for the case the dao does not hold all the rewards of the time of the call .
and there was also a need to maintain the flexibility to call each one of it separately.

contracts/schemes/Competition.sol
orenyodfat: - in this case all the suggestions which has the same number of votes will get their split reward divided by their count.
- for this we are maintaining votesToSuggestions

orenyodfat:
thanks @ben-kaufman @jellegerbrandy @dkent600 for comments and feedback.
I think I responds to most of your comment. It indeed a complex one.
I will go and merge it in order to unblock the upper layers.
Please submit issues for any issue /concerns/questions re this pr.

PR #704, opened 2020-01-07
Added AuthorizedMintRep scheme

### LINKED ISSUE #703: add AuthorizedMintRep scheme
Needed for ETHDenver dao.
A scheme which whitelist an account for minting rep.
initialised with:
-  activationStartTime
-  activationEndTime
-  total reputation reward allowed (0 value will allow infinite reputation reward).
 method :
 - reputationMint(address, amount) onlyAuthorized.

ben-kaufman:
Fix #703

### DIFF
--- a/contracts/schemes/AuthorizedMintRep.sol
+++ b/contracts/schemes/AuthorizedMintRep.sol
@@ -1,0 +1,75 @@
+pragma solidity 0.5.13;
+
+import "openzeppelin-solidity/contracts/math/SafeMath.sol";
+import "../controller/Controller.sol";
+
+/**
+ * @title A scheme for reputation allocation according to token balances
+ *        This contract is assuming that the token contract is paused, and one cannot transfer its tokens.
+ */
+
+contract AuthorizedMintRep {
+    using SafeMath for uint256;
+
+    Avatar public avatar;
+    uint256 public activationStartTime;
+    uint256 public activationEndTime;
+    uint256 public repRewardLeft;
+    address public authorizedAddress;
+    bool public limitRepReward;
+
+    /**
+     * @dev Throws if called by an unauthorized account.
+     */
+    modifier onlyAuthorized() {
+        require(msg.sender == authorizedAddress, "Caller is not authorized");
+        _;
+    }
+
+    /**
+     * @dev initialize
+     * @param _avatar the avatar to mint reputation from
+     * @param _activationStartTime start time for allowing minting
+     * @param _activationEndTime end time for allowing minting
+     * @param _maxRepReward maximum reputation mintable by this scheme
+     * @param _authorizedAddress address authorized for minting reputation
+     */
+    function initialize(
+        Avatar _avatar,
+        uint256 _activationStartTime,
+        uint256 _activationEndTime,
+        uint256 _maxRepReward,
+        address _authorizedAddress
+    ) external {
+        require(avatar == Avatar(0), "can be called only one time");
+        require(_avatar != Avatar(0), "avatar cannot be zero");
+        require(_activationStartTime < _activationEndTime, "_activationStartTime < _activationEndTime");
+        avatar = _avatar;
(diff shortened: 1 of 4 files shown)


contracts/schemes/AuthorizedMintRep.sol
```diff
@@ -0,0 +1,75 @@
+pragma solidity 0.5.13;
+
+import "openzeppelin-solidity/contracts/math/SafeMath.sol";
+import "../controller/Controller.sol";
+
+/**
+ * @title A scheme for reputation allocation according to token balances
```
orenyodfat: this is not right.

contracts/schemes/AuthorizedMintRep.sol
```diff
@@ -0,0 +1,75 @@
...
+    uint256 public activationStartTime;
+    uint256 public activationEndTime;
+    uint256 public repRewardLeft;
+    address public authorizedAddress;
+    bool public limitRepReward;
+
+    /**
+     * @dev Throws if called by an unauthorized account.
+     */
+    modifier onlyAuthorized() {
```
orenyodfat: how about use ownable ?

orenyodfat: and onlyOwner

contracts/schemes/AuthorizedMintRep.sol
ben-kaufman: fixed

contracts/schemes/AuthorizedMintRep.sol
```diff
@@ -0,0 +1,75 @@
...
+    ) external {
+        require(avatar == Avatar(0), "can be called only one time");
+        require(_avatar != Avatar(0), "avatar cannot be zero");
+        require(_activationStartTime < _activationEndTime, "_activationStartTime < _activationEndTime");
+        avatar = _avatar;
+        activationStartTime = _activationStartTime;
+        activationEndTime = _activationEndTime;
+        repRewardLeft = _maxRepReward;
+        authorizedAddress = _authorizedAddress;
+        limitRepReward = _maxRepReward != 0;
```
orenyodfat: what does it means != 0 ?

contracts/schemes/AuthorizedMintRep.sol
```diff
@@ -0,0 +1,75 @@
...
+    }
+
+    /**
+     * @dev reputationMint function
+     * @param _beneficiary the beneficiary address to redeem for
```
orenyodfat: redeem ?

contracts/schemes/AuthorizedMintRep.sol
```diff
@@ -0,0 +1,75 @@
...
+     * @param _amount the agreementHash hash
```
orenyodfat: _agreementHash ?

contracts/schemes/AuthorizedMintRep.sol
```diff
@@ -0,0 +1,75 @@
...
+     */
+    function reputationMint(address _beneficiary, uint256 _amount) external onlyAuthorized {
```
orenyodfat: does this function should return bool same as the reputation mint ?

contracts/schemes/AuthorizedMintRep.sol
ben-kaufman: If it is not 0 I want to skip the check of the limit of reputation

contracts/schemes/AuthorizedMintRep.sol
ben-kaufman: No, but it can if we want it to.

contracts/schemes/AuthorizedMintRep.sol
orenyodfat: this is strange syntax

ben-kaufman: Yes, what's weird about this syntax?

PR #717, opened 2020-03-12
Create SchemeFactory

ben-kaufman:


contracts/schemes/SchemeFactory.sol
```diff
@@ -0,0 +1,172 @@
...
+import "@daostack/infra-experimental/contracts/votingMachines/IntVoteInterface.sol";
+import "@daostack/infra-experimental/contracts/votingMachines/VotingMachineCallbacksInterface.sol";
+import "../votingMachines/VotingMachineCallbacks.sol";
+import "../utils/DAOFactory.sol";
+import "@openzeppelin/upgrades/contracts/Initializable.sol";
+
+
+/**
+ * @title A factory and registrar for Schemes for organizations
+ * @dev The SchemeFactory is used for deploying and registering schemes at organizations
```
orenyodfat: The SchemeFactory is used for deploying and registering schemes to organisations

contracts/schemes/SchemeFactory.sol
```diff
@@ -0,0 +1,172 @@
...
+    // a SchemeProposal is a  proposal to add or remove a scheme to/from the an organization
+    struct SchemeProposal {
+        string schemeName;
+        bytes schemeData;
+        uint64[3] packageVersion;
+        address schemeToReplace;
+        bytes4 permissions;
+    }
+
+    mapping(bytes32=>SchemeProposal) public organizationProposals;
```
orenyodfat: organizationProposals -> proposals

contracts/schemes/SchemeFactory.sol
```diff
@@ -0,0 +1,172 @@
...
+        votingMachine = _votingMachine;
+        voteParams = _voteParams;
+        daoFactory = _daoFactory;
+    }
+
+    /**
+    * @dev execution of proposals, can only be called by the voting machine in which the vote is held.
+    * @param _proposalId the ID of the voting in the voting machine
+    * @param _decision the voting result, 1 yes and 2 is no.
+    */
```
orenyodfat: @return ?

contracts/schemes/SchemeFactory.sol
```diff
@@ -0,0 +1,172 @@
...
+        bytes4 _permissions,
+        address _schemeToReplace,
+        string _descriptionHash
+    );
+
+    event ProposalExecuted(address indexed _avatar, bytes32 indexed _proposalId, int256 _param);
+    event ProposalDeleted(address indexed _avatar, bytes32 indexed _proposalId);
+
+    // a SchemeProposal is a  proposal to add or remove a scheme to/from the an organization
+    struct SchemeProposal {
```
orenyodfat: Proposal

contracts/schemes/SchemeFactory.sol
```diff
@@ -0,0 +1,172 @@
...
+    onlyVotingMachine(_proposalId)
+    returns(bool) {
+        SchemeProposal memory proposal = organizationProposals[_proposalId];
+        require(bytes(proposal.schemeName).length != 0);
+        delete organizationProposals[_proposalId];
+        emit ProposalDeleted(address(avatar), _proposalId);
+        if (_decision == 1) {
+            // Define controller and get the params:
+            Controller controller = Controller(avatar.owner());
+
```
orenyodfat: maybe check if proposal.schemeName != "" to enable only unregister proposal

contracts/schemes/SchemeFactory.sol
```diff
@@ -0,0 +1,172 @@
...
+    /**
+    * @dev create a proposal to register a scheme
+    * @param _packageVersion the Arc version to use for deploying the scheme
+    * @param _schemeName the name of the scheme to be added
+    * @param _schemeData initialize data for the scheme to be added
+    * @param _permissions the permission of the scheme to be registered
+    * @param _schemeToReplace address of scheme to replace with the new scheme (zero for none)
+    * @param _descriptionHash proposal's description hash
+    * @return a proposal Id
+    * @dev NB: not only proposes the vote, but also votes for it
```
orenyodfat: please remove this

contracts/schemes/SchemeFactory.sol
```diff
@@ -0,0 +1,172 @@
...
+        bytes schemeData;
+        uint64[3] packageVersion;
+        address schemeToReplace;
+        bytes4 permissions;
+    }
+
+    mapping(bytes32=>SchemeProposal) public organizationProposals;
+
+    IntVoteInterface public votingMachine;
+    bytes32 public voteParams;
```
orenyodfat: so the params to remove are the same as param to register ?

test/schemefactory.js
```diff
@@ -0,0 +1,140 @@
...
+        //Vote with reputation to trigger execution
+        var proposalId = await helpers.getValueFromLogs(tx, '_proposalId',1);
+
+         //Vote with reputation to trigger execution
+         tx = await testSetup.schemeFactoryParams.votingMachine.absoluteVote.vote(proposalId,2,0,helpers.NULL_ADDRESS,{from:accounts[2]});
+         //should not register because the decision is "no"
+         let proxyEvents = await registration.daoFactory.getPastEvents("ProxyCreated", {fromBlock: tx.receipt.blockNumber, toBlock: tx.receipt.blockNumber});
+         assert.equal(proxyEvents.length,0);//proposalType
+        });
+});
```
orenyodfat: any test to only remove scheme ?

contracts/schemes/SchemeFactory.sol
```diff
@@ -0,0 +1,172 @@
...
+import "../votingMachines/VotingMachineCallbacks.sol";
+import "../utils/DAOFactory.sol";
+import "@openzeppelin/upgrades/contracts/Initializable.sol";
+
+
+/**
+ * @title A factory and registrar for Schemes for organizations
+ * @dev The SchemeFactory is used for deploying and registering schemes at organizations
+ */
+
```
orenyodfat: this line space is redundant

contracts/schemes/SchemeFactory.sol
ben-kaufman: There is no @return also in the other schemes documenting it, it either returns true if the function didn't fail and revert if it did fail, but it can only return true...

contracts/schemes/SchemeFactory.sol
ben-kaufman: As we want to do everything in a single proposal, we can only have one params. We can't have one proposal voted on with 2 sets of params

test/schemefactory.js
ben-kaufman: I didn't allow this functionality, will add it.

ben-kaufman: added

contracts/utils/DAOFactory.sol
```diff
@@ -20,7 +20,7 @@ contract DAOFactory is Initializable {
         address indexed _reputation,
         address _daotoken
     );
-    
+
```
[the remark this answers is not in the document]
ben-kaufman: The linter was warning on this so I removed the whitespace. Beter like that

contracts/schemes/SchemeFactory.sol
ben-kaufman: added

contracts/schemes/SchemeFactory.sol
orenyodfat: so please add that to the function documentation :)

ben-kaufman: ok

contracts/schemes/SchemeFactory.sol
```diff
@@ -0,0 +1,173 @@
...
+    function executeProposal(bytes32 _proposalId, int256 _decision)
+    external
+    onlyVotingMachine(_proposalId)
+    returns(bool) {
+        Proposal memory proposal = proposals[_proposalId];
+        delete proposals[_proposalId];
+        emit ProposalDeleted(address(avatar), _proposalId);
+        Controller controller = Controller(avatar.owner());
+        if (_decision == 1) {
+            if (bytes(proposal.schemeName).length > 0) {
```
orenyodfat: does proposal.schemeName != "" does not work ?

contracts/schemes/SchemeFactory.sol
```diff
@@ -0,0 +1,173 @@
...
+        bytes memory _schemeData,
+        bytes4 _permissions,
+        address _schemeToReplace,
+        string memory _descriptionHash
+    )
+    public
+    returns(bytes32)
+    {
+        if (bytes(_schemeName).length > 0) {
+            // propose
```
orenyodfat: redundant /not in place

contracts/schemes/SchemeFactory.sol
ben-kaufman: no

contracts/schemes/SchemeFactory.sol
```diff
@@ -0,0 +1,172 @@
...
+                                        proposal.packageVersion,
+                                        proposal.schemeName,
+                                        address(avatar),
+                                        proposal.schemeData));
+
+                require(controller.registerScheme(scheme, proposal.permissions), "faild to register new scheme");
+            }
+
+
+            if (proposal.schemeToReplace != address(0) && controller.isSchemeRegistered(proposal.schemeToReplace)) {
```
orenyodfat: controller.isSchemeRegistered is checked already at line 131 ?

ben-kaufman: I'm checking here again because it is possible that meanwhile, as the proposal was in voting phase, the scheme was already removed by another proposal/scheme.

ben-kaufman: It's not a must I guess, but could be a good check to have

orenyodfat: I am not sure that on this case we want to revert ..though.
I think it is better that the execution will not revert as it might do a problem with the gov process (e.g the last vote which cause execution).

orenyodfat: please consider not emitting the ProposalExecuted event or emit another event ..as it might confused the upper layers

ben-kaufman: It doesn't cause a revert here though. It just won't run the unregister

ben-kaufman: This if here is actually necessary to prevent a revert if the scheme is not registered anymore for some reason. Without it, the next line will revert (as the unregister will return false)

PR #718, opened 2020-03-18
Add ArcUpgradeScheme

ben-kaufman:


contracts/schemes/ArcUpgradeScheme.sol
```diff
@@ -0,0 +1,200 @@
...
+
+        emit ProposalExecutedByVotingMachine(address(avatar), _proposalId, _decision);
+        return true;
+    }
+
+    /**
+    * @dev execution of proposals after it has been decided by the voting machine
+    * @param _proposalId the ID of the voting in the voting machine
+    */
+    function execute(bytes32 _proposalId) public {
```
orenyodfat: why this is public function ?
is there a chance that the genericCall will revert on the execution time and will not refer after it ?

contracts/schemes/ArcUpgradeScheme.sol
```diff
@@ -0,0 +1,200 @@
...
+    function proposeUpgrade(
+        uint64[3] memory _packageVersion,
+        bytes32[] memory _contractsNames,
+        address[] memory _contractsToUpgrade,
+        string memory _descriptionHash)
+    public
+    returns(bytes32)
+    {
+        require(package.hasVersion(_packageVersion), "Specified version doesn't exist in the Package");
+        for (uint256 i = 0; i < _contractsToUpgrade.length; i++) {
```
orenyodfat: I think we should get the max possible length which can be executed gas wised and cap the length of this array with that

contracts/schemes/ArcUpgradeScheme.sol
```diff
@@ -0,0 +1,200 @@
...
+            proposalId,
+            _packageVersion,
+            _contractsNames,
+            _contractsToUpgrade,
+            _descriptionHash
+        );
+        return proposalId;
+    }
+
+    function bytes32ToStr(bytes32 x) private pure returns (string memory) {
```
orenyodfat: this bit complex util function is also at https://github.com/daostack/arc/blob/arc-factory/contracts/utils/DAOFactory.sol#L234 .
how about make it in a lib and use it in both places ?

contracts/schemes/ArcUpgradeScheme.sol
```diff
@@ -0,0 +1,200 @@
...
+import "@openzeppelin/upgrades/contracts/Initializable.sol";
+import "@openzeppelin/upgrades/contracts/application/Package.sol";
+import "@openzeppelin/upgrades/contracts/application/ImplementationProvider.sol";
+
+
+/**
+ * @title ArcUpgradeScheme.
+ * @dev  A scheme for proposing updating the versions DAOs contracts
+ */
+contract ArcUpgradeScheme is VotingMachineCallbacks, ProposalExecuteInterface, Initializable {
```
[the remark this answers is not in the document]
ben-kaufman: We already have UpgradeScheme

contracts/schemes/ArcUpgradeScheme.sol
ben-kaufman: Only chance I think is if the Avatar is not the admin of the contract proposed. Not sure it’s necessary to have this public but I just sticked to the same way you made GenericScheme

contracts/schemes/ArcUpgradeScheme.sol
ben-kaufman: 20 seems to be around the max, so will cap it there ok?

contracts/schemes/ArcUpgradeScheme.sol
```diff
@@ -0,0 +1,200 @@
...
+    event ProposalExecutedByVotingMachine(
+        address indexed _avatar,
+        bytes32 indexed _proposalId,
+        int256 _param
+    );
+
+    event ProposalDeleted(address indexed _avatar, bytes32 indexed _proposalId);
+
+    // Details of a voting proposal:
+    struct UpgradeProposal {
```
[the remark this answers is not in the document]
ben-kaufman: fixed

contracts/schemes/ArcUpgradeScheme.sol
```diff
@@ -0,0 +1,200 @@
...
+import "@daostack/infra-experimental/contracts/votingMachines/VotingMachineCallbacksInterface.sol";
+import "../votingMachines/VotingMachineCallbacks.sol";
+import "@openzeppelin/upgrades/contracts/Initializable.sol";
+import "@openzeppelin/upgrades/contracts/application/Package.sol";
+import "@openzeppelin/upgrades/contracts/application/ImplementationProvider.sol";
+
+
+/**
+ * @title ArcUpgradeScheme.
+ * @dev  A scheme for proposing updating the versions DAOs contracts
```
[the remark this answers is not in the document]
ben-kaufman: fixed

contracts/schemes/ArcUpgradeScheme.sol
ben-kaufman: moved to a lib

contracts/schemes/ArcUpgradeScheme.sol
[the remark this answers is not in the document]
ben-kaufman: Fixed

contracts/schemes/ArcUpgradeScheme.sol
orenyodfat: it should not be public

test/upgradescheme.js
```diff
@@ -1,239 +1,444 @@
...
+            web3.utils.fromAscii("Avatar"),web3.utils.fromAscii("DAOToken"),web3.utils.fromAscii("Reputation"),
+            web3.utils.fromAscii("Avatar"),web3.utils.fromAscii("DAOToken"),web3.utils.fromAscii("Reputation"),
+            web3.utils.fromAscii("Avatar"),web3.utils.fromAscii("DAOToken"),web3.utils.fromAscii("Reputation"),
+            web3.utils.fromAscii("Avatar"),web3.utils.fromAscii("DAOToken"),web3.utils.fromAscii("Reputation"),
+            web3.utils.fromAscii("Avatar"),web3.utils.fromAscii("DAOToken"),web3.utils.fromAscii("Reputation"),
+            web3.utils.fromAscii("Avatar"),web3.utils.fromAscii("DAOToken"),web3.utils.fromAscii("Reputation"),
+            web3.utils.fromAscii("Avatar"),web3.utils.fromAscii("DAOToken"),web3.utils.fromAscii("Reputation"),
+            web3.utils.fromAscii("Avatar"),web3.utils.fromAscii("DAOToken"),web3.utils.fromAscii("Reputation")
+          ],
+          [
```
orenyodfat: how about filling this array in a for loop ?

test/upgradescheme.js
```diff
@@ -1,239 +1,444 @@
...
+          web3.utils.fromAscii("Avatar"),web3.utils.fromAscii("DAOToken"),web3.utils.fromAscii("Reputation"),
+          web3.utils.fromAscii("Avatar"),web3.utils.fromAscii("DAOToken"),web3.utils.fromAscii("Reputation"),
+          web3.utils.fromAscii("Avatar"),web3.utils.fromAscii("DAOToken"),web3.utils.fromAscii("Reputation"),
+          web3.utils.fromAscii("Avatar"),web3.utils.fromAscii("DAOToken"),web3.utils.fromAscii("Reputation"),
+          web3.utils.fromAscii("Avatar"),web3.utils.fromAscii("DAOToken"),web3.utils.fromAscii("Reputation"),
+          web3.utils.fromAscii("Avatar"),web3.utils.fromAscii("DAOToken"),web3.utils.fromAscii("Reputation"),
+          web3.utils.fromAscii("Avatar"),web3.utils.fromAscii("DAOToken"),web3.utils.fromAscii("Reputation")
+        ],
+        [
+          testSetup.org.avatar.address, testSetup.org.token.address, testSetup.org.reputation.address,
```
orenyodfat: for loop ?

test/upgradescheme.js
ben-kaufman: done

test/upgradescheme.js
ben-kaufman: done

contracts/schemes/ArcUpgradeScheme.sol
ben-kaufman: You mean here or also in the GenericScheme?

orenyodfat: here. 
in GenericScheme it is different case where there is a chance that the target contract call will revert though the arc tx will not due to the nature of contract.call ...

The case in this scheme is different as there is no way the target contract can revert.

contracts/schemes/ArcUpgradeScheme.sol
orenyodfat: does the number 60 verified/tested ?

ben-kaufman: It is tested that 60 works. I think more is possible theoretically but not much so just rounded it down.

contracts/schemes/ArcUpgradeScheme.sol
ben-kaufman: ok.

test/upgradescheme.js
```diff
@@ -196,8 +196,8 @@ contract('UpgradeScheme', function(accounts) {
       let contractsNames = [];
       let contractsToUpgrade = [];
       for (let i = 0; i < 21; i++) {
```
orenyodfat: how about make it 60 ? like in the contract limit

ben-kaufman: the loop adds 3 at a time

contracts/schemes/UpgradeScheme.sol
```diff
@@ -61,106 +73,106 @@ contract UpgradeScheme is Initializable, VotingMachineCallbacks, ProposalExecute
...
-                bytes4 permissions = controller.schemesPermissions(address(this));
-                require(
-                controller.registerScheme(proposal.upgradeContract, permissions)
+    function executeProposal(bytes32 _proposalId, int256 _decision)
+    external
+    onlyVotingMachine(_proposalId)
+    returns(bool) {
+        Proposal storage proposal = organizationProposals[_proposalId];
+        require(proposal.exist, "must be a live proposal");
+        require(proposal.passed == false, "cannot execute twice");
```
orenyodfat: do we now need these ?

contracts/schemes/UpgradeScheme.sol
```diff
@@ -61,106 +73,106 @@ contract UpgradeScheme is Initializable, VotingMachineCallbacks, ProposalExecute
...
+                    abi.encodeWithSignature("upgradeTo(address)", updatedImp),
+                    0
                 );
-                if (proposal.upgradeContract != address(this)) {
-                    require(controller.unregisterSelf());
-                }
             }
+
+            delete organizationProposals[_proposalId];
+            emit ProposalDeleted(address(avatar), _proposalId);
```
orenyodfat: this can be taken out of the if statement ..

contracts/schemes/UpgradeScheme.sol
```diff
@@ -61,106 +73,106 @@ contract UpgradeScheme is Initializable, VotingMachineCallbacks, ProposalExecute
...
+            emit ProposalExecuted(address(avatar), _proposalId);
+        } else {
+            delete organizationProposals[_proposalId];
+            emit ProposalDeleted(address(avatar), _proposalId);
```
orenyodfat: same as above

contracts/schemes/UpgradeScheme.sol
```diff
@@ -61,106 +73,106 @@ contract UpgradeScheme is Initializable, VotingMachineCallbacks, ProposalExecute
...
                 );
-                if (proposal.upgradeContract != address(this)) {
-                    require(controller.unregisterSelf());
-                }
             }
+
+            delete organizationProposals[_proposalId];
+            emit ProposalDeleted(address(avatar), _proposalId);
+            emit ProposalExecuted(address(avatar), _proposalId);
+        } else {
```
orenyodfat: no need else

contracts/schemes/UpgradeScheme.sol
ben-kaufman: Exists I think yes, but I removed the passed property

dOrgJelli:
Amazing! :clap: :heart_eyes:

contracts/schemes/UpgradeScheme.sol
```diff
@@ -61,106 +66,99 @@ contract UpgradeScheme is Initializable, VotingMachineCallbacks, ProposalExecute
...
-                controller.registerScheme(proposal.upgradeContract, permissions)
+    function executeProposal(bytes32 _proposalId, int256 _decision)
+    external
+    onlyVotingMachine(_proposalId)
+    returns(bool) {
+        Proposal storage proposal = organizationProposals[_proposalId];
+        require(proposal.exist, "must be a live proposal");
+
+        if (_decision == 1) {
+            proposal.exist = false;
```
orenyodfat: so this is another protection ..along with onlyVotingMachine .. let it be :)

PR #720, opened 2020-03-22
Remove use of package keyword

ben-kaufman:


contracts/schemes/UpgradeScheme.sol
```diff
@@ -27,7 +27,8 @@ contract UpgradeScheme is VotingMachineCallbacks, ProposalExecuteInterface, Init
 
     event ProposalExecuted(
         address indexed _avatar,
-        bytes32 indexed _proposalId
+        bytes32 indexed _proposalId,
+        int256 _param
```
orenyodfat: why is this needed ?
should it be a boolean ?

ben-kaufman: It makes things slightly easier in the subgraph and we have it in all other contracts like that.

orenyodfat: should it be a boolean ?
should it be called _decision ?

ben-kaufman: Not sure, it's an int everywhere else, I think the voting machine returns int so that's why.

orenyodfat: this are due to history reasons , we now have the chance to rewrite history :) 
I think it should be boolean , name `decision` .
does subgraph look for this event ? why ?

ben-kaufman: to mark the proposal as executed and in this case also update the contract info versions.

ben-kaufman: changed

PR #723, opened 2020-03-24
add JoinAndQuit scheme

orenyodfat:


orenyodfat:
- missing function :  if the deadline pass and the goal has not been reached... members should  be able to  redeemed their funding . it is not exactly the same as rageQuit .
@jellegerbrandy   what do you think will be a proper name for this function ?

contracts/controller/Avatar.sol
```diff
@@ -169,5 +170,16 @@ contract Avatar is Initializable, Ownable {
         return true;
     }
 
+    /**
+    * @dev setDB set a key value in the dao db
+    * @param _key a string
+    * @param _value a string
+    * @return _value which represents a success
```
jellegerbrandy: @return true if successful

contracts/controller/Avatar.sol
```diff
@@ -169,5 +170,16 @@ contract Avatar is Initializable, Ownable {
...
+    */
+    function setDB(string calldata _key, string calldata _value) external onlyOwner returns(bool) {
```
jellegerbrandy: `setDBValue` maybe?

contracts/libs/StringUtil.sol
```diff
@@ -0,0 +1,14 @@
...
+
+
+library StringUtil {
+
+    function hashCompareWithLengthCheck(string memory _a, string memory _b) internal returns (bool) {
+
+        if (bytes(_a).length != bytes(_b).length) {
+            return false;
+        } else {
+            return keccak256(abi.encodePacked(_a)) == keccak256(abi.encodePacked(_b));
```
jellegerbrandy: I'm proabably missing a point, but is this not equivalent to `abi.encodePacked(_a) == abi.encodePacked(_b)`?

orenyodfat: the compiler does not like it :
`TypeError: Operator == not compatible with types bytes memory and bytes memory`

contracts/schemes/JoinAndQuit.sol
```diff
@@ -0,0 +1,243 @@
...
+     * @param _minFeeToJoin minimum fee required to join
+     * @param _memberReputation the repution which will be allocated for members
+     * @param _fundingGoal the funding goal
+     * @param _fundingGoalDeadLine the funding goal deadline
+     */
+    function initialize(
+        Avatar _avatar,
+        IntVoteInterface _votingMachine,
+        bytes32 _voteParams,
+        IERC20 _fundingToken,
```
jellegerbrandy: This means we will not be able to fund with ETH? 

I am not sure if XDAI on the poa network is an ERC20 token (I believe not, but I'll need to check)

contracts/schemes/JoinAndQuit.sol
```diff
@@ -0,0 +1,243 @@
...
+    mapping(bytes32=>Proposal) public proposals;
+    mapping(address=>uint256) public fundings;
+
+    IntVoteInterface public votingMachine;
+    bytes32 public voteParams;
+    Avatar public avatar;
+    IERC20 public fundingToken;
+    uint256 public minFeeToJoin;
+    uint256 public memberReputation;
+    uint256 public fundingGoal;
```
jellegerbrandy: Other schemes may want to access this data. With the current logic, they can only see if the goal has been reached, but not what the goal is.  It would perhaps be better to strore these setting on the Avatar.db, just to not close any options of other schemes... (same with minFee, etc)

contracts/schemes/JoinAndQuit.sol
orenyodfat: `This means we will not be able to fund with ETH?`
Is that a requirement ? I never saw that in the spec .

- xdai is erc677 which is erc20 compatible

contracts/schemes/JoinAndQuit.sol
```diff
@@ -0,0 +1,243 @@
...
+    )
+    external
+    initializer
+    {
+        require(_avatar != Avatar(0), "avatar cannot be zero");
+        avatar = _avatar;
+        votingMachine = _votingMachine;
+        voteParams = _voteParams;
+        fundingToken = _fundingToken;
+        minFeeToJoin = _minFeeToJoin;
```
jellegerbrandy: If a DAO wanted to change this parameter, how would that work? Do we want to add a function to change these values? Or would that mean to remove this scheme and deploy a new one with different params?

contracts/schemes/JoinAndQuit.sol
orenyodfat: yes . maybe. for now there is no such a scheme req and we will use the db as a way to share data between schemes only at runtime setup ..not as a parameters setup on deployment time. though we can do that if there will be a real need.

contracts/schemes/JoinAndQuit.sol
```diff
@@ -0,0 +1,243 @@
...
+    using SafeMath for uint;
+    using SafeERC20 for address;
+    using StringUtil for string;
+
+    event JoinInProposal(
+        address indexed _avatar,
+        bytes32 indexed _proposalId,
+        string _descriptionHash,
+        address _proposedMember,
+        uint256 _fundAmount
```
jellegerbrandy: We are calling it "fee" when we write "minFeeToJoin". Let's be consistent about this. 

I think "funding" is a cbit confusing here..  here it is about  money flowing into the DAO, but we are also using the term for money flowing OUT of the DAO (as in "funding request"). 

We can use "fee" - so instead of `_fundAmont` just write `_fee` or `_feeAmount`.

contracts/schemes/JoinAndQuit.sol
orenyodfat: Yes.
using [SchemeFactory](https://github.com/daostack/arc/blob/arc-factory/contracts/schemes/SchemeFactory.sol) scheme .

contracts/schemes/JoinAndQuit.sol
```diff
@@ -0,0 +1,243 @@
...
+    */
+    function checkFundedBeforeDeadLine() private {
+        if ((avatar.db(FUNDED_BEFORE_DEADLINE_KEY).hashCompareWithLengthCheck(FUNDED_BEFORE_DEADLINE_VALUE) == false) &&
+            (fundingToken.balanceOf(address(avatar)) >= fundingGoal) &&
+            // solhint-disable-next-line not-rely-on-time
+            (now < fundingGoalDeadLine)) {
+            require(
+            Controller(
+            avatar.owner()).setDBValue("FUNDED_BEFORE_DEADLINE", "TRUE"));
+            emit FundedDeadLineReached(address(avatar));
```
jellegerbrandy: This may very well happen way before the funding deadline is reached, so this event name is confusing.

`DAOIsFunded` would be better, or even `FundedBeforeDeadline` to be consistent with the db key.

contracts/schemes/JoinAndQuit.sol
```diff
@@ -0,0 +1,243 @@
...
+    * @dev checkFundedBeforeDeadLine check if funding goal reached.
+    */
+    function checkFundedBeforeDeadLine() private {
+        if ((avatar.db(FUNDED_BEFORE_DEADLINE_KEY).hashCompareWithLengthCheck(FUNDED_BEFORE_DEADLINE_VALUE) == false) &&
+            (fundingToken.balanceOf(address(avatar)) >= fundingGoal) &&
+            // solhint-disable-next-line not-rely-on-time
+            (now < fundingGoalDeadLine)) {
+            require(
+            Controller(
+            avatar.owner()).setDBValue("FUNDED_BEFORE_DEADLINE", "TRUE"));
```
jellegerbrandy: you want to use the variable FUNDED_BEFORE_DEADLINE_KEY here, I think

contracts/schemes/JoinAndQuit.sol
jellegerbrandy: Or "donataion" ,as in

`minDonationToJoin`
`donations` --> the mapping that is now called `fundings`
`_donationAmount`

contracts/schemes/JoinAndQuit.sol
```diff
@@ -0,0 +1,244 @@
...
+        Controller(
+        avatar.owner()).externalTokenTransfer(fundingToken, msg.sender, refund));
+        emit RageQuit(address(avatar), refund);
+    }
+
+    /**
+    * @dev checkFundedBeforeDeadLine check if funding goal reached.
+    */
+    function checkFundedBeforeDeadLine() private {
+        if ((avatar.db(FUNDED_BEFORE_DEADLINE_KEY).hashCompareWithLengthCheck(FUNDED_BEFORE_DEADLINE_VALUE) == false) &&
```
jellegerbrandy: why is this so complex? Why not just compare with the empty string?

contracts/schemes/JoinAndQuit.sol
```diff
@@ -0,0 +1,244 @@
...
+            (fundingToken.balanceOf(address(avatar)) >= fundingGoal) &&
```
jellegerbrandy: An alternative would be to check if `totalDonation >= fundingGoal`

orenyodfat: not sure re that . as someone can send funds directly to the avatar.

contracts/schemes/JoinAndQuit.sol
orenyodfat: not so complex

contracts/schemes/JoinAndQuit.sol
```diff
@@ -0,0 +1,244 @@
...
+    * @dev rageQuit quit from the dao.
+    * can be done on any time
+    * REFUND = USER_DONATION * CURRENT_DAO_BALANCE / TOTAL_DONATIONS
+    * @return refund the refund amount
+    */
+    function rageQuit() public returns(uint256 refund) {
+        require(fundings[msg.sender] > 0, "no fund to RageQuit");
+        uint256 userDonation = fundings[msg.sender];
+        fundings[msg.sender] = 0;
+        refund = userDonation.mul(fundingToken.balanceOf(address(avatar))).div(totalDonation);
```
jellegerbrandy: So I have a doubt here but no particular solution.

There is a scenario here in which a rage quitter will get more money out then it put it. Namely if someone sends $1.000.000 to the DAO directly, because it is a nice cause...  it is not correct that a rage quitter ha the right to a "proportional part" of this money...

contracts/schemes/JoinAndQuit.sol
```diff
@@ -0,0 +1,244 @@
...
+    }
+
+    /**
+    * @dev Submit a proposal for to join in a dao
+    * @param _descriptionHash A hash of the proposal's description
+    * @param _feeAmount - the amount to fund the dao with. should be >= the minimum fee to join
+    * @param _member the member to join in - if this address is zero the msg.sender will be set as the member
+    * @return proposalId the proposal id
+    */
+    function proposeToJoinIn(
```
jellegerbrandy: make it `proposeToJoin`

contracts/schemes/JoinAndQuit.sol
orenyodfat: yes. this is a valid case. if the dao have some earning and the member is part of it ..so why not ..

contracts/schemes/JoinAndQuit.sol
```diff
@@ -0,0 +1,244 @@
...
+        string memory _descriptionHash,
+        uint256 _feeAmount,
+        address _member
```
jellegerbrandy: maybe call this `_proposedMember` to be consistent with tthe Event parameter?

contracts/schemes/JoinAndQuit.sol
jellegerbrandy: Funding with ETH is not particularly a requirement (it would be nice, but we do not need that). But funding with XDAI is definitely something we want to do :-)

And as far as I undstand (but again, I am not 100% sure), xdai is the "native token" on the xdai network, meaning it is like ETH on the normal chain, and not an ERC20 contract...

contracts/schemes/JoinAndQuit.sol
```diff
@@ -0,0 +1,244 @@
...
+    function executeProposal(bytes32 _proposalId, int256 _decision)
+    external
+    onlyVotingMachine(_proposalId)
+    returns(bool) {
+        require(proposals[_proposalId].accepted == false);
+        require(proposals[_proposalId].proposedMember != address(0));
+        Proposal memory proposal = proposals[_proposalId];
+        // Check if vote was successful:
+        if (_decision == 1) {
+            proposals[_proposalId].accepted = true;
```
jellegerbrandy: meaning `proposals[_proposalId].accepted` could have been false before, when the decision was already taking? Where do you see that? 

Maybe `accepted` here menas `executed`?

contracts/schemes/JoinAndQuit.sol
```diff
@@ -0,0 +1,244 @@
...
+        refund = userDonation.mul(fundingToken.balanceOf(address(avatar))).div(totalDonation);
+        totalDonation = totalDonation.sub(userDonation);
+        require(
+        Controller(
+        avatar.owner()).externalTokenTransfer(fundingToken, msg.sender, refund));
+        emit RageQuit(address(avatar), refund);
+    }
+
+    /**
+    * @dev checkFundedBeforeDeadLine check if funding goal reached.
```
jellegerbrandy: This does a bit more though than just check, though - it sets the db value
Call it `setFundingGoalReachedFlag` (it's ugly but at least it says what is does)

contracts/schemes/JoinAndQuit.sol
```diff
@@ -0,0 +1,244 @@
...
+        });
+        return proposalId;
+    }
+
+    /**
+    * @dev RedeemReputation reward for proposal
+    * @param _proposalId the ID of the voting in the voting machine
+    * @return reputation the redeemed reputation.
+    */
+    function redeemReputation(bytes32 _proposalId) public returns(uint256 reputation) {
```
jellegerbrandy: Must this be a separate function? What about making this a part of the `executeProposal` function?

contracts/schemes/JoinAndQuit.sol
orenyodfat: yes , it is false before.
this is for the proposal decision. 
the proposal can be executed though not accepted by the dao .

contracts/schemes/JoinAndQuit.sol
orenyodfat: we can do that though :
1. the `withdraw pattern` is more secure. https://ethereum.stackexchange.com/questions/67342/what-is-the-withdrawal-pattern 
2. do we want the last voter for the case of abs vote or the executer will pay the gas ? 

so why not sending the funds on the `withdraw pattern` ? good question .we might need that  that immediately upon the proposal execution /dao decision.

contracts/schemes/JoinAndQuit.sol
```diff
@@ -0,0 +1,244 @@
...
+    using SafeMath for uint;
+    using SafeERC20 for address;
+    using StringUtil for string;
+
+    event JoinInProposal(
+        address indexed _avatar,
+        bytes32 indexed _proposalId,
+        string _descriptionHash,
+        address _proposedMember,
+        uint256 _feeAmount
```
dkent600: I like the word "donation" a lot better than "fee"

PR #725, opened 2020-03-25
Add FundingRequest scheme

ben-kaufman:


contracts/schemes/FundingRequest.sol
```diff
@@ -0,0 +1,158 @@
...
+    struct Proposal {
+        address beneficiary;
+        uint256 amount;
+        string descriptionHash;
+        uint256 executionTime;
+    }
+
+    mapping(bytes32=>Proposal) public proposals;
+
+    string public constant FUNDED_BEFORE_DEADLINE_KEY = "FUNDED_BEFORE_DEADLINE";
```
orenyodfat: having that in 2 places is not so good
does import these values from the JoinAndQuit scheme might work ?
I think it is ..
or lets create an interface file which both scheme can import and use .

contracts/schemes/FundingRequest.sol
```diff
@@ -0,0 +1,158 @@
...
+    string public constant FUNDED_BEFORE_DEADLINE_VALUE = "TRUE";
+    IntVoteInterface public votingMachine;
+    bytes32 public voteParams;
+    Avatar public avatar;
+    IERC20 public fundingToken;
```
orenyodfat: we need to support also funding in eth . see [JoinAndQuit](https://github.com/daostack/arc/blob/arc-factory/contracts/schemes/JoinAndQuit.sol#L79) scheme.

contracts/schemes/FundingRequest.sol
```diff
@@ -0,0 +1,158 @@
...
+        address _beneficiary,
+        uint256 _amount,
+        string memory _descriptionHash
+    )
+    public
+    returns(bytes32)
+    {
+        require(
+            avatar.db(FUNDED_BEFORE_DEADLINE_KEY).hashCompareWithLengthCheck(FUNDED_BEFORE_DEADLINE_VALUE),
+            "funding from this scheme was not yet approved"
```
orenyodfat: funding is not allowed yet

contracts/schemes/FundingRequest.sol
ben-kaufman: Updated

contracts/schemes/FundingRequest.sol
ben-kaufman: Added

contracts/schemes/FundingRequest.sol
ben-kaufman: fixed

contracts/schemes/FundingRequest.sol
```diff
@@ -0,0 +1,164 @@
...
+        string descriptionHash;
+        uint256 executionTime;
+    }
+
+    mapping(bytes32=>Proposal) public proposals;
+
+    IntVoteInterface public votingMachine;
+    bytes32 public voteParams;
+    Avatar public avatar;
+    address public fundingToken;
```
orenyodfat: maybe type should be IERC20 . see [here](https://github.com/daostack/arc/blob/arc-factory/contracts/schemes/JoinAndQuit.sol#L67)

contracts/schemes/FundingRequest.sol
```diff
@@ -0,0 +1,164 @@
...
+    * @dev Redeem proposal funding
+    * @param _proposalId the ID of the voting in the voting machine
+    */
+    function redeem(bytes32 _proposalId) public {
+        Proposal memory _proposal = proposals[_proposalId];
+        Proposal storage proposal = proposals[_proposalId];
+        require(proposal.executionTime != 0);
+        proposal.executionTime = 0;
+        if (fundingToken == address(0)) {
+            require(Controller(avatar.owner()).sendEther(_proposal.amount, _proposal.beneficiary));
```
orenyodfat: require msg error ?

contracts/schemes/FundingRequest.sol
```diff
@@ -0,0 +1,164 @@
...
+        } else {
+            require(Controller(avatar.owner()).externalTokenTransfer(
```
orenyodfat: require msg error ?

PR #727, opened 2020-03-26
Common schemes fix

orenyodfat:
- remove Donation event
- FundingRequest lint 
- msg.sender is proposedMember

### DIFF
--- a/contracts/schemes/JoinAndQuit.sol
+++ b/contracts/schemes/JoinAndQuit.sol
@@ -35,6 +35,7 @@ contract JoinAndQuit is
 
     event RageQuit(
         address indexed _avatar,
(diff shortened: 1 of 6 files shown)


contracts/schemes/JoinAndQuit.sol
```diff
@@ -33,13 +33,9 @@ contract JoinAndQuit is
...
     );
 
-    event Donation(
-        address indexed _avatar,
-        uint256 indexed _donation
-    );
-
     event RageQuit(
         address indexed _avatar,
+        address indexed _rageQuiter,
```
ben-kaufman: `_rageQuiter` -> `rageQuitter `

PR #750, opened 2020-05-11
Separate Vault from Avatar

ben-kaufman:


contracts/controller/Avatar.sol
```diff
@@ -90,6 +90,7 @@ contract Avatar is Initializable, Ownable {
     returns(bool success, bytes memory returnValue) {
         if (_value > 0) {
             vault.sendEther(_value, address(this));
+            emit SendEther(_value, _contract);
```
orenyodfat, line 93: lets move this event to the vault

PR #752, opened 2020-05-19
add dictator.sol

orenyodfat:


contracts/schemes/Dictator.sol
```diff
@@ -0,0 +1,36 @@
...
+        //register a scheme with full permission :)
+        require(Controller(avatar.owner()).registerScheme(_scheme, 0x0000001f), "Fail to register scheme");
+    }
+
+    /**
+     * @dev _initialize
+     * @param _avatar the scheme avatar
+     * @param _owner the contract owner
+     */
+    function initialize(Avatar _avatar, address _owner) public initializer {
```
ben-kaufman, line 30: why not register the owner as admin right away in the initialize and then reregister self?

orenyodfat: because the initialize function is called as the process of instance the scheme before the scheme is registered .

ben-kaufman: oh right, so nevermind

dkent600:
Would be great if docs like this could be in the code:

https://github.com/daostack/arc/pull/752#pullrequestreview-416941211

orenyodfat:
> * deploy our daos with the dictator scheme registered, and with dictator.owner set to 0xdaostack
> * call dictator.registerScheme(0xdaostack)
> * and now we can do everything the controller allows us to do from the 0xdaostack account

not exactly :
- The owner of dictator is 0xdaostack .
- 0xdaostack can register any arbitrary scheme or account with full permission to the dao by calling call dictator.registerScheme(0xanyarbitraryschemeoraccount)

0xanyarbitraryschemeoraccount can do what ever.

changing ownership of Dictator will be 1 step : 
 - dictator.transferOwnership(0xnewOwner)

Such a contract is needed to register 0xanyarbitraryschemeoraccount due to the fact that daoFactory can register contracts only.

PR #753, opened 2020-05-19
Common token

orenyodfat:
An ERC20 token with an admin account that can arbitrarily mint and burn tokens in each account.

### DIFF
--- a/contracts/misc/CommonToken.sol
+++ b/contracts/misc/CommonToken.sol
@@ -1,0 +1,21 @@
+pragma solidity ^0.5.17;
+
+import "@openzeppelin/contracts-ethereum-package/contracts/token/ERC20/StandaloneERC20.sol";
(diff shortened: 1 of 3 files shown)


contracts/misc/CommonToken.sol
```diff
@@ -0,0 +1,21 @@
...
+ * @dev ERC20 compatible token. It is a mintable, burnable token(onlyOwner)
+ */
+contract CommonToken is StandaloneERC20 {
+
+    /**
+     * @dev Destroys `amount` tokens from the caller.
+     *
+     * See {ERC20-_burn}.
+     */
+    function burn(uint256 amount) public onlyMinter {
```
jellegerbrandy, line 18: I think we need: 
```burn(address account, uint256 amount) public onlyMinter {
  _burn(account, amount)
```

dkent600:
Would be great to have a description in each new contract PR of the purpose and motivation behind the new contract.  Otherwise one has to read the code and guess.

PR #754, opened 2020-05-19
dxdaoTimeLock

orenyodfat:


contracts/misc/TimeLocking.sol
```diff
@@ -0,0 +1,26 @@
...
+
+    function () external payable {
+    }
+
+    function withdraw() external {
+        require(msg.sender == owner, "only owner can withdraw");
+       // solhint-disable-next-line not-rely-on-time
+        require(releaseTime < now, "cannot withdraw before releaseTime");
+          // solhint-disable-next-line avoid-call-value
+        (bool success, ) = owner.call.value(address(this).balance)("");
```
AugustoL, line 22: Why not `msg.sender.transfer(address(this).balance)` ?

orenyodfat: because transfer is limited in gas. it might work for dxdao avatar though no promise for other contracts which consume gas on the fallback function.
https://diligence.consensys.net/blog/2019/09/stop-using-soliditys-transfer-now/

contracts/test/Wallet.sol
```diff
@@ -27,4 +27,12 @@ contract Wallet is Ownable {
         emit Pay(_beneficiary, amount);
     }
 
+    function genericCall(address _contract, bytes memory _encodedABI)
+    public
+    returns(bool success, bytes memory returnValue) {
+       // solhint-disable-next-line avoid-low-level-calls
+        (success, returnValue) = _contract.call(_encodedABI);
```
AugustoL, line 34: Why instead of sending the _encodeABI we execute the withdraw function with the signature that we already know it will have?
```
(success, returnValue) = _contract.call(abi.encodeWithSignature("withdraw()"));
```

Doing this we make sure that the ONLY function that can be executed is the `withdraw()`

AugustoL:
> Why is this contract name prefixed with "dxDao"? Does it have no possible more widespread use?

yeah, right? I guess the idea is to have this feature to all DAOs..

contracts/test/Wallet.sol
orenyodfat: It is a public generic function which can call any contract specific Abis.
this is just a test helper contract and is not part of any dao.

orenyodfat:
> > Why is this contract name prefixed with "dxDao"? Does it have no possible more widespread use?
> 
> yeah, right? I guess the idea is to have this feature to all DAOs..

I am not sure yet . We might have it specific to dxdao by hard coded its address and release time.though might complicate the tests.
yea. lets make it general

PR #756, opened 2020-05-22
etherGC.sol

orenyodfat:


dkent600:
Can you spell out "GC"?  No idea what it means...

contracts/globalConstraints/EtherGC.sol
```diff
@@ -0,0 +1,84 @@
...
+     * @dev check the allowance of ether sent per period.
+     * @return bool which represents a success
+     */
+    function post(address, bytes32, bytes32) public returns(bool) {
+        require(msg.sender == controller, "only controller is authorize to call");
+        uint256 currentPeriodIndex = (block.number - startBlock)/constraintPeriodWindow;
+        if ((block.number - startBlock) % constraintPeriodWindow > 0) {
+            currentPeriodIndex = currentPeriodIndex.add(1);
+        }
+        if (avatarBalanceBefore >= avatar.balance) {
```
jellegerbrandy, line 67: so you are checking the netto balance, not how much ether the avatar actually did send (i.e. if the constraint is 1 ETH/day, then it would be ok if the avatar receives 100.000 and sends 99.999)

contracts/globalConstraints/EtherGC.sol
```diff
@@ -0,0 +1,84 @@
...
+            uint256 ethSentAmount = avatarBalanceBefore.sub(avatar.balance);
+            totalAmountSentForPeriods[currentPeriodIndex] =
+            totalAmountSentForPeriods[currentPeriodIndex].add(ethSentAmount);
+            require(totalAmountSentForPeriods[currentPeriodIndex] <= amountAllowedPerPeriod,
+            "exceed the amount allowed");
```
jellegerbrandy, line 72: maybe at a reference to where this error message comes from, something like `Violation of Global constraint EtherGC: amount sent exceed in current period°

contracts/globalConstraints/EtherGC.sol
```diff
@@ -0,0 +1,84 @@
...
+
+    /**
+     * @dev check the allowance of ether sent per period.
+     * @return bool which represents a success
+     */
+    function post(address, bytes32, bytes32) public returns(bool) {
+        require(msg.sender == controller, "only controller is authorize to call");
+        uint256 currentPeriodIndex = (block.number - startBlock)/constraintPeriodWindow;
+        if ((block.number - startBlock) % constraintPeriodWindow > 0) {
+            currentPeriodIndex = currentPeriodIndex.add(1);
```
jellegerbrandy, line 65: why do we need this? Was line 53 not enough?

contracts/globalConstraints/EtherGC.sol
```diff
@@ -0,0 +1,84 @@
...
+     * @return true
+     */
+    function pre(address, bytes32, bytes32) public returns(bool) {
+        require(msg.sender == controller, "only controller is authorize to call");
+        avatarBalanceBefore = avatar.balance;
+        return true;
+    }
+
+    /**
+     * @dev check the allowance of ether sent per period.
```
jellegerbrandy, line 58: "and throws an error if the constraint is violated"

test/etherGC.js
```diff
@@ -0,0 +1,75 @@
...
+      assert.equal(await etherGC.avatar(),avatar.address);
+      assert.equal(await etherGC.amountAllowedPerPeriod(),web3.utils.toWei('5', "ether"));
+      assert.equal(await etherGC.constraintPeriodWindow(),10);
+    });
+
+  it("send ether check", async () => {
+
+    await setup();
+    try {
+     await etherGC.initialize(avatar.address,20,web3.utils.toWei('5', "ether"),controller.address); //10 blocks ,5 eth
```
jellegerbrandy, line 44: comment says 10 blocks, code says 20

contracts/globalConstraints/EtherGC.sol
```diff
@@ -0,0 +1,84 @@
...
+
+import "openzeppelin-solidity/contracts/token/ERC20/IERC20.sol";
+import "./GlobalConstraintInterface.sol";
+import "openzeppelin-solidity/contracts/math/SafeMath.sol";
+
+
+/**
+ * @title EtherGC ether constraint per period
+ */
+contract EtherGC {
```
jellegerbrandy, line 11: you have a GlobalConstraint interface to inherit from?

contracts/globalConstraints/EtherGC.sol
```diff
@@ -0,0 +1,84 @@
...
+   * @dev initialize
+   * @param _avatar the avatar to mint reputation from
+   * @param _constraintPeriodWindow the constraintPeriodWindow in blocks units
+   * @param _amountAllowedPerPeriod the amount of eth to constraint for each period
+   */
+    function initialize(
+        address _avatar,
+        uint256 _constraintPeriodWindow,
+        uint256 _amountAllowedPerPeriod,
+        address _controller
```
jellegerbrandy, line 33: why do you need the controller as a separate argument, can't you get that from the avatar_?

test/etherGC.js
```diff
@@ -0,0 +1,75 @@
...
+      //increment 21 blocks in ganache
+      await Reputation.new();
+
+    }
+
+    await controller.sendEther(web3.utils.toWei('1', "ether"), accounts[2],avatar.address);
+    await controller.sendEther(web3.utils.toWei('4', "ether"), accounts[2],avatar.address);
+    //send more than 10 eth
+    await web3.eth.sendTransaction({from:accounts[0],to:avatar.address, value: web3.utils.toWei('10', "ether")});
+    await controller.sendEther(web3.utils.toWei('4', "ether"), accounts[2],avatar.address);
```
jellegerbrandy, line 72: I think you need to explain in a comment why this statement is not failing. Afaics, if this and the previous lines fall within the same period, you have sent 9 ether and violated the constraint. So by some coincidence, this last transfer happens in the next period?

contracts/globalConstraints/EtherGC.sol
orenyodfat: this will not be allowed by this constraint .

contracts/globalConstraints/EtherGC.sol
orenyodfat: e.g because. solidity div operation is floor rounded.

contracts/globalConstraints/EtherGC.sol
orenyodfat: it comes from this contract . not sure we need to explicit say where is come from.
but why not

contracts/globalConstraints/EtherGC.sol
```diff
@@ -0,0 +1,78 @@
...
+
+
+/**
+ * @title EtherGC ether constraint per period
+ */
+contract EtherGC is GlobalConstraintInterface {
+    using SafeMath for uint256;
+
+    uint256 public periodLength; //the period length in blocks units
+    uint256 public amountAllowedPerPeriod;
```
leviadam, line 16: Not sure if this should be percentage-wise, or absolute values. Probably better to keep it absolute.

contracts/globalConstraints/EtherGC.sol
```diff
@@ -0,0 +1,78 @@
...
+     * @return bool which represents a success
+     */
+    function post(address, bytes32, bytes32) public returns(bool) {
+        require(msg.sender == avatar.owner(), "only avatar owner is authorize to call");
+        if (avatarBalanceBefore > address(avatar).balance) {
+            uint256 currentPeriodIndex = (block.number - startBlock)/periodLength;
+            totalAmountSentPerPeriod[currentPeriodIndex] =
+            totalAmountSentPerPeriod[currentPeriodIndex].add(avatarBalanceBefore.sub(address(avatar).balance));
+            require(totalAmountSentPerPeriod[currentPeriodIndex] <= amountAllowedPerPeriod,
+            "Violation of Global constraint EtherGC:amount sent exceed in current period");
```
leviadam, line 66: Should set back avatarBalanceBefore to zero, as it will save gas.

contracts/globalConstraints/ReputationGC.sol
```diff
@@ -0,0 +1,96 @@
...
+import "../controller/Avatar.sol";
+
+
+/**
+ * @title ReputationGC reputation mint/butn constraint per period
+ */
+contract ReputationGC is GlobalConstraintInterface {
+    using SafeMath for uint256;
+
+    uint256 public periodLength; //the period length in blocks units
```
leviadam, line 15: Same as in EtherGC

contracts/schemes/GlobalConstraintAddOrRemove.sol
```diff
@@ -0,0 +1,53 @@
...
+
+import "../controller/Controller.sol";
+
+/**
+ * @title A scheme for adding or removing a global constraint
+ * The scheme will unregister itself after register the globalConstraint
+ * This scheme should be register to the dao with permission 0x00000004
+ */
+
+contract GlobalConstraintAddOrRemove {
```
ben-kaufman, line 11: What's this contract for? We already have `GlobalConstraintRegistrar` which does this much better...

PR #757, opened 2020-05-31
None universal voting machines

orenyodfat:
-DAOFactory :  DAO creation in a single transaction .
- None universal voting machines .
- Use infra exp 1.1.0-rc.18


contracts/schemes/GlobalConstraintRegistrar.sol
```diff
@@ -29,39 +29,48 @@ contract GlobalConstraintRegistrar is VotingMachineCallbacks, ProposalExecuteInt
...
-        super._initializeGovernance(_avatar, _votingMachine, _voteParamsHash, _votingParams, _voteOnBehalf);
+        super._initializeGovernance(
+            _avatar,
+            _votingParams,
+            _voteOnBehalf,
+            _daoFactory,
+            _stakingToken,
+            address(this),
+            address(this),
+            address(this),
```
ben-kaufman, line 71: If in all our schemes it's all the scheme address 3 times why not just make a `scheme` parameter for all the 3 instead?

Anyway this seems wrong actually, shouldn't this be the Avatar address, then twice the scheme address, instead of 3 times the scheme?

This comment is for all schemes...

contracts/schemes/JoinAndQuit.sol
```diff
@@ -66,51 +67,73 @@ contract JoinAndQuit is
...
+
+        (
+            address  fundingToken,
+            uint256  minFeeToJoin,
+            uint256  memberReputation,
+            uint256  fundingGoal,
+            uint256  fundingGoalDeadline,
+            bool  rageQuitEnable
+        )  =
+        abi.decode(_encodedJoinAndQuitParams, (address, uint256, uint256, uint256, uint256, bool));
```
ben-kaufman, line 127: Why is that? Not that I see an issue, just want to understand why the added complexity here?

contracts/schemes/VoteInOrganizationScheme.sol
```diff
@@ -34,21 +34,35 @@ contract VoteInOrganizationScheme is VotingMachineCallbacks, ProposalExecuteInte
...
      */
     function initialize(
         Avatar _avatar,
-        IntVoteInterface _votingMachine,
         uint256[11] calldata _votingParams,
         address _voteOnBehalf,
-        bytes32 _voteParamsHash
+        DAOFactory _daoFactory,
+        address _stakingToken,
+        uint64[3] calldata _packageVersion,
```
ben-kaufman, line 50: Maybe we could take that from the Avatar or somewhere else? I don't really like how this is duplicated in all the code... (Also relevant for daofactory)

contracts/utils/DAOFactory.sol
```diff
@@ -40,35 +41,81 @@ contract DAOFactory is Initializable {
     mapping(address=>Locks) public locks;
     App public app;
     string public constant PACKAGE_NAME = "DAOstack";
-    //this is here due to "stack too deep issue"
-    uint64[3] private packageVersion;
 
+    // orgName - The name of the new organization
```
ben-kaufman, line 45: Delete if not using this

package.json
```diff
@@ -69,7 +70,7 @@
   },
   "homepage": "https://daostack.io",
   "dependencies": {
-    "@daostack/infra-experimental": "0.0.1-rc.16",
```
ben-kaufman, line 72: Let's also bump arc version

contracts/schemes/ArcScheme.sol
```diff
@@ -24,35 +32,62 @@ contract ArcScheme is Initializable {
...
     function _initializeGovernance(
         Avatar _avatar,
-        IntVoteInterface _votingMachine,
-        bytes32 _voteParamsHash,
         uint256[11] memory _votingParams,
-        address _voteOnBehalf
+        address _voteOnBehalf,
+        DAOFactory _daoFactory,
+        address _stakingToken,
+        address _organization,
```
ben-kaufman, line 51: Why is there both `_organization ` and `_avatar`?
Shouldn't we just have `_avatar `?

contracts/schemes/ArcScheme.sol
```diff
@@ -24,35 +32,62 @@ contract ArcScheme is Initializable {
...
+        address _callbacks,
+        address _authorizedToPropose,
```
ben-kaufman, line 53: Maybe combine here `_callbacks` with `_authorizedToPropose `?
This is anyway only for our schemes and I don't think we have a use case which needs them to be different.

contracts/schemes/JoinAndQuit.sol
orenyodfat: this is needed to solve stack too deep issue.

contracts/schemes/VoteInOrganizationScheme.sol
orenyodfat: the alternative is to always use lasted package for the voting machines.
need to give it more thoughts.

contracts/schemes/ArcScheme.sol
orenyodfat: there might be a use case for that . 
for now lets keep it like that.

ben-kaufman: ok

contracts/utils/DAOFactory.sol
```diff
@@ -193,42 +218,52 @@ contract DAOFactory is Initializable {
...
      * @param _avatar organization avatar (returns from forgeOrg)
-     * @param _schemesNames the schemes name to register for the organization
-     * @param _schemesData the schemes initilization data
-     * @param _schemesInitilizeDataLens the schemes initilization data lens (at _schemesData)
-     * @param _permissions the schemes permissions.
-     * @param _metaData dao meta data hash
+     * @param _encodedSetSchemesParams _setSchemes parameters
+     * @param _packageVersion package version
      */
     function _setSchemes (
```
ben-kaufman, line 224: Each scheme receives DAOFactory, and package version in its init params, which creates a large duplication of data. It might be too complicated to do but it would be better if we could instead of passing it to each scheme we could pass it here for the schemes

PR #758, opened 2020-06-01
feat(contracts): Add WalletScheme contract and tests

AugustoL:
Adds the wallet scheme contract, it is based in the genericScheme but it doesnt make the genericCall
to the controller, instead it can execte multiple staticCalls to any contract. (It does not allow
execute staticCalls on the scheme itself and controller).

### DIFF
--- a/contracts/schemes/WalletScheme.sol
+++ b/contracts/schemes/WalletScheme.sol
@@ -1,0 +1,179 @@
+pragma solidity 0.5.17;
+pragma experimental ABIEncoderV2;
+
+import "@daostack/infra/contracts/votingMachines/IntVoteInterface.sol";
+import "@daostack/infra/contracts/votingMachines/VotingMachineCallbacksInterface.sol";
+import "../votingMachines/VotingMachineCallbacks.sol";
+
+
+/**
+ * @title WalletScheme.
+ * @dev  A scheme for proposing and executing calls to any contract except itself and controller
+ */
+contract WalletScheme is VotingMachineCallbacks, ProposalExecuteInterface {
+    event NewCallProposal(
+        address[] _to,
+        bytes32 indexed _proposalId,
+        bytes[]   _callData,
+        uint256[] _value,
+        string  _descriptionHash
+    );
+
+    event ProposalExecuted(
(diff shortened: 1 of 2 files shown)


orenyodfat:
@AugustoL could you please explain more about the motivation behind this scheme .
In relation to security concerns : 
- This scheme has no constraint re the target contract it sends the funds to(which are part of the proposal) so in a sense you are getting more general porpose scheme though you are paying with   more vulnerability.

AugustoL:
> @AugustoL could you please explain more about the motivation behind this scheme .
> In relation to security concerns :
> 
> * This scheme has no constraint re the target contract it sends the funds to(which are part of the proposal) so in a sense you are getting more general porpose scheme though you are paying with   more vulnerability.

Hola @orenyodfat, effectively. The idea of the WalletScheme is to have the flexibility to make static calls directly to any contract.
The idea is to have multiple wallets with different level of security (weekly withdrawal limit, minimum execution time, required reputation, etc). 

By not calling the controller we wont be able to use the global constraints, right? maybe we would have to add this configuration in the wallet scheme itself, right?

orenyodfat:
> > @AugustoL could you please explain more about the motivation behind this scheme .
> > In relation to security concerns :
> > 
> > * This scheme has no constraint re the target contract it sends the funds to(which are part of the proposal) so in a sense you are getting more general porpose scheme though you are paying with   more vulnerability.
> 
> Hola @orenyodfat, effectively. The idea of the WalletScheme is to have the flexibility to make static calls directly to any contract.
> The idea is to have multiple wallets with different level of security (weekly withdrawal limit, minimum execution time, required reputation, etc).
> 
> By not calling the controller we wont be able to use the global constraints, right? maybe we would have to add this configuration in the wallet scheme itself, right?

Yes. no global constraint on this scheme.

Few options for that :
1. multiple standard genericSchemes. each for each wallet.
2. use ContributionReward proposals to send funds from the avatar to each wallet.
3. have a generic scheme with white listed target contracts.
Please note that option 3 is new (same as the one you implemented) and will require support on the higher layers (migration/subgraph/arc.js/alchemy).
option 1&2 are already supported by alchemy .

PR #760, opened 2020-06-03
Add NFTManager + tests

tspoff:


tspoff:
@orenyodfat My JS test file is formatted differently than the rest - is there a contribution / style guide for formatting?

contracts/misc/NFTManager.sol
```diff
@@ -0,0 +1,19 @@
...
+    function initialize() public initializer {
+        Ownable.initialize(msg.sender);
+    }
+
+    /// @notice Transfer specified NFT to a given recipient
+    /// @param recipient Address to transfer NFT to
+    /// @param nftContract NFT token contract to transfer from (e.g. Cryptokitties contract)
+    /// @param tokenId ID of specific NFT to transfer
+    function sendNFT(address recipient, IERC721 nftContract, uint tokenId) external onlyOwner {
+        nftContract.transferFrom(address(this), recipient, tokenId);
```
orenyodfat, line 17: should we use safeTransferFrom . see https://github.com/OpenZeppelin/openzeppelin-contracts/blob/master/contracts/token/ERC721/IERC721.sol#L59 .
might even better to use https://github.com/OpenZeppelin/openzeppelin-contracts/blob/master/contracts/token/ERC721/IERC721.sol#L128 as it gives more options to pass data along the call.

orenyodfat:
lets make travis green.

contracts/misc/NFTManager.sol
tspoff: My issue with safeTransferFrom() is the requirement that receiving contracts must implement IERC721Reciever - there may be a number of contract recipients that don’t implement this.

However, I agree safeTransfer should be the default option unless there’s a specific reason not to use it.

Here's my proposed change, with explicit functions for sending data or no data with safeTransfer(): https://github.com/dOrgTech/arc/pull/1/files

orenyodfat: please pr your proposal to this repo.

contracts/misc/NFTManager.sol
```diff
@@ -0,0 +1,63 @@
...
+
+    function initialize() public initializer {
+        Ownable.initialize(msg.sender);
+    }
+
+    /// @notice Safe Transfer specified NFT to a given recipient
+    /// @param recipient Address to transfer NFT to
+    /// @param nftContract NFT token contract to transfer from (e.g. Cryptokitties contract)
+    /// @param tokenId ID of specific NFT to transfer
+    function sendNFT(address recipient, IERC721 nftContract, uint tokenId) public onlyOwner {
```
orenyodfat, line 19: this can be `external`

orenyodfat: why this function needed at all ? does sendNFTWithData is not enough ?

contracts/misc/NFTManager.sol
```diff
@@ -0,0 +1,63 @@
...
+     * returned can be obtained as `this.onERC721Received.selector`. This
+     * function MAY throw to revert and reject the transfer.
+     * Note: the ERC721 contract address is always the message sender.
+     * @param operator The address which called `safeTransferFrom` function
+     * @param from The address which previously owned the token
+     * @param tokenId The NFT identifier which is being transferred
+     * @param data Additional data with no specified format
+     * @return bytes4 `bytes4(keccak256("onERC721Received(address,address,uint256,bytes)"))`
+     */
+    function onERC721Received(address operator, address from, uint256 tokenId, bytes memory data)
```
orenyodfat, line 57: please remove unused var name ..
`function onERC721Received(address, address, uint256, bytes )`

test/nftmanager.js
```diff
@@ -82,11 +82,20 @@ const setup = async function(
     account2 = reputationAccount;
   }
   testSetup.proxyAdmin = accounts[5];
+  console.log(
```
orenyodfat, line 85: why console.log ?

test/nftmanager.js
```diff
@@ -0,0 +1,689 @@
...
+  var testSetup = new helpers.TestSetup();
+  testSetup.standardTokenMock = await ERC20Mock.new(accounts[1], 100);
+  registration = await helpers.registerImplementation();
+  testSetup.reputationArray = [20, 10, 70];
+  var account2;
+  if (reputationAccount === 0) {
+    account2 = accounts[2];
+  } else {
+    account2 = reputationAccount;
+  }
```
orenyodfat, line 83: this is needed for. something ?

PR #763, opened 2020-06-10
Return universal VMs, leave DAOFactory changes

ben-kaufman:


contracts/schemes/SignalScheme.sol
```diff
@@ -30,43 +30,41 @@ contract SignalScheme is VotingMachineCallbacks, ProposalExecuteInterface {
         bool executed;
     }
 
-    uint256 public signalType;
+    struct Parameters {
+        bytes32 voteApproveParams;
+        IntVoteInterface intVote;
+        uint256 signalType;
+        Avatar avatar;
```
orenyodfat, line 37: no need avatar here.

PR #769, opened 2020-07-09
Add Transition Scheme

ben-kaufman:


contracts/schemes/TransitionScheme.sol
```diff
@@ -0,0 +1,88 @@
...
+
+    /**
+     * @dev transferAssets function
+     * transfer the DAO assets to a new DAO
+     */
+    function transferAssets() external {
+        for (uint256 i=0; i < assetAddresses.length; i++) {
+            Controller(avatar.owner()).genericCall(
+                assetAddresses[i],
+                abi.encodeWithSelector(selectors[i], newAvatar),
```
orenyodfat, line 50: is that assume that the selector to transfer always get one param (address) ?

ben-kaufman: Yes. I didn't want to allow any data, though it shouldn't be a security risk so long as it is properly verified when doing the initialize

contracts/schemes/TransitionScheme.sol
```diff
@@ -0,0 +1,88 @@
...
+                avatar,
+                0
```
orenyodfat, line 52: emit an event Transfer(_avatar,_asset,newAvatar)

contracts/schemes/TransitionScheme.sol
```diff
@@ -0,0 +1,88 @@
...
+        selectors = _selectors;
+    }
+
+    /**
+     * @dev transferAssets function
+     * transfer the DAO assets to a new DAO
+     */
+    function transferAssets() external {
+        for (uint256 i=0; i < assetAddresses.length; i++) {
+            Controller(avatar.owner()).genericCall(
```
orenyodfat, line 48: get the return value

contracts/schemes/TransitionScheme.sol
ben-kaufman: The contract transferred will emit a transfer only if succeeded anyway, why emit also here?

contracts/schemes/TransitionScheme.sol
ben-kaufman: What is the point of that, we have nothing to do with the returned value here...

contracts/schemes/TransitionScheme.sol
orenyodfat: 1. it will be easy to monitor which assets where transfer
2. not all assets gurrenty to emit events when transfer

contracts/schemes/TransitionScheme.sol
orenyodfat: check if the transfer success or not .

dkent600:
What is the function of this scheme?

ben-kaufman:
To help DAOs migrate from the old stack to the new one.

contracts/schemes/TransitionScheme.sol
dkent600: ... which one could test in advance using `.call`

contracts/schemes/TransitionScheme.sol
```diff
@@ -0,0 +1,88 @@
...
+    address payable public newAvatar;
+    address[] public externalTokens;
+    address[] public assetAddresses;
+    bytes4[] public selectors;
+
+    /**
+     * @dev initialize
+     * @param _avatar the avatar to migrate from
+     * @param _newAvatar the avatar to migrate to
+     * @param _externalTokens external tokens to allow transfer to the new avatar
```
dkent600, line 23: Why is _externakTokens not just to be included in `_assetAddresses` + `_selectors`?

contracts/schemes/TransitionScheme.sol
```diff
@@ -0,0 +1,88 @@
...
+     * @param _assetAddresses the assets to transfer
+     * @param _selectors the functions to call to to transfer the assets
```
dkent600, line 25: Does this include reputation?

dkent600, line 25: Would be good to document here the form a function must take, like what parameters and return value.

ben-kaufman: No, reputation is recreated in the new DAO

ben-kaufman: It is a standard selector per the ABI structure.

contracts/schemes/TransitionScheme.sol
ben-kaufman: I don't think I understand the question. Tokens contracts are not owned by the DAO, only amounts of tokens, so it's a different function.

contracts/schemes/TransitionScheme.sol
orenyodfat: no

contracts/schemes/TransitionScheme.sol
```diff
@@ -0,0 +1,97 @@
...
+     */
+    function initialize(
+        Avatar _avatar,
+        address payable _newAvatar,
+        address[] calldata _externalTokens,
+        address[] calldata _assetAddresses,
```
orenyodfat, line 31: how about cap the number of asserts with a number which is applicable ?

ben-kaufman: Why? The number of assets is determined in the initialized, so the DAO knows how many and which assets will be transferred before adding the scheme.

contracts/schemes/TransitionScheme.sol
```diff
@@ -0,0 +1,97 @@
...
+    address[] public assetAddresses;
+    bytes4[] public selectors;
+
+    /**
+     * @dev initialize
+     * @param _avatar the avatar to migrate from
+     * @param _newAvatar the avatar to migrate to
+     * @param _externalTokens external tokens to allow transfer to the new avatar
+     * @param _assetAddresses the assets to transfer
+     * @param _selectors the functions to call to to transfer the assets
```
orenyodfat, line 25: how about passing here selectors as strings so it will be easy to verify.

contracts/schemes/TransitionScheme.sol
orenyodfat: I mean in terms of gas usage ..while iterating over the assets.

ben-kaufman: I think we can do enough assets at a time to make it unnecessary, but if you want we can add it. I could add a test to see how many we can do at a time but it'd be just a rough estimation as different assets will probably cost differently.

orenyodfat: test with normal transfer ownership

contracts/schemes/TransitionScheme.sol
dkent600: ok, that answers the question.

contracts/schemes/TransitionScheme.sol
dkent600: Is there any other plan for transferring rep from a 1.0 DAO to a 2.0 DAO?

dkent600: A selector is an ABI specification for any arbitrary function.  Don't the selectors in this list have to specify a function that has a specific set of parameters and parameter types?

orenyodfat: sure.

contracts/schemes/TransitionScheme.sol
ben-kaufman: ok I added a test transferring 101 assets.

contracts/schemes/TransitionScheme.sol
```diff
@@ -0,0 +1,97 @@
...
+        address[] calldata _assetAddresses,
+        bytes4[] calldata _selectors
+    ) external {
+        require(_assetAddresses.length == _selectors.length, "Arrays length mismatch");
+        require(avatar == Avatar(0), "can be called only one time");
+        require(_avatar != Avatar(0), "avatar cannot be zero");
+        avatar = _avatar;
+        newAvatar = _newAvatar;
+        externalTokens = _externalTokens;
+        assetAddresses = _assetAddresses;
```
orenyodfat, line 40: lets cap this array length with 100

contracts/schemes/TransitionScheme.sol
```diff
@@ -0,0 +1,100 @@
...
+
+import "../controller/Controller.sol";
+
+/**
+ * @title A scheme for transitioning the DAO's assets to a new one.
+ */
+
+contract TransitionScheme {
+
+    uint public constant ASSETS_CAP = 100;
```
orenyodfat, line 11: uint -> uint256

PR #770, opened 2020-07-15
Competition : Add Redemption period

orenyodfat:
- this is to avoid the dao depend on all suggester to redeem their rewards to get back the leftover using sendLeftOverFunds


contracts/schemes/Competition.sol
```diff
@@ -274,13 +274,9 @@ contract Competition {
     */
     function sendLeftOverFunds(bytes32 _proposalId) public {
         // solhint-disable-next-line not-rely-on-time
-        require(proposals[_proposalId].endTime < now, "competition is still on");
+        require(proposals[_proposalId].endTime.add(REDEMPTION_PERIOD) < now, "redemption period is till on");
```
ben-kaufman, line 277: typo: `till` -> `still`

PR #774, opened 2020-07-16
Add TokenTrade scheme

ben-kaufman:


contracts/schemes/TokenTrade.sol
```diff
@@ -0,0 +1,159 @@
+pragma solidity ^0.6.10;
```
orenyodfat, line 1: lets make is solidity 0.6.11

contracts/schemes/TokenTrade.sol
```diff
@@ -0,0 +1,159 @@
...
+
+
+/**
+ * @title A scheme for join in a dao.
+ * - A member can be proposed to join in by sending a min amount of fee.
+ * - A member can ask to quite (RageQuit) a dao on any time.
+ * - A member can donate to a dao.
+ */
+contract TokenTrade is VotingMachineCallbacks, ProposalExecuteInterface {
+    using SafeMath for uint;
```
orenyodfat, line 14: uint -> unit256

contracts/schemes/TokenTrade.sol
```diff
@@ -0,0 +1,159 @@
...
+    /**
+     * @dev initialize
+     * @param _avatar the avatar this scheme referring to.
+     * @param _votingMachine the voting machines address to
+     * @param _votingParams genesisProtocol parameters - valid only if _voteParamsHash is zero
+     * @param _voteOnBehalf genesisProtocol parameter - valid only if _voteParamsHash is zero
+     * @param _voteParamsHash voting machine parameters.
+     */
+    function initialize(
+        Avatar _avatar,
```
orenyodfat, line 59: I think it should get also the tradingToken

contracts/schemes/TokenTrade.sol
ben-kaufman: Seems problematic to push just one file version... maybe push all in a separate PR?

contracts/schemes/TokenTrade.sol
ben-kaufman: Why not allow any token? What's the reason to limit it?

contracts/schemes/TokenTrade.sol
orenyodfat: ok.

contracts/schemes/TokenTrade.sol
orenyodfat: 1. easy in the UI.
2. simple code
3. more secure

contracts/schemes/TokenTrade.sol
```diff
@@ -0,0 +1,159 @@
...
+    )
+    public
+    returns(bytes32)
+    {
+        require(
+            address(_sendToken) != address(0) && address(_receiveToken) != address(0),
+            "Token address must not be null"
+        );
+        require(_sendTokenAmount > 0 && _receiveTokenAmount > 0, "Token amount must be greater than 0");
+
```
orenyodfat, line 132: I think we need a way to `lock` the recieve token in the scheme.. so there will not be a case where the proposal execute and the dao cannot pay .
so maybe upon proposal the scheme will ask the receive token amount from the avatar.. ,if not enough exist will revert.
if proposal is accepted the user will redeem (from the scheme) .
if proposal is rejected the allocated token will be sent back to the avatar

ben-kaufman: This is bad because an attacker can open dummy proposal and lock the avatar's balance. The way I handled that case is that if upon execution the DAO doesn't have enough funds, the proposal will be canceled as if it was rejected, and the user will get his tokens back.

contracts/schemes/TokenTrade.sol
```diff
@@ -0,0 +1,179 @@
...
+    event ProposalExecuted(address indexed _avatar, bytes32 indexed _proposalId, int256 _decision);
+
+    struct Proposal {
+        address beneficiary;
+        IERC20 sendToken;
+        uint256 sendTokenAmount;
+        IERC20 receiveToken;
+        uint256 receiveTokenAmount;
+        bool exist;
```
orenyodfat, line 46: does the `exist` var is really needed  ?
does checking other filed like `beneficiary` or `sendToken` is not enough ?

contracts/schemes/TokenTrade.sol
```diff
@@ -0,0 +1,179 @@
...
+    returns(bytes32)
+    {
+        require(
+            address(_sendToken) != address(0) && address(_receiveToken) != address(0),
+            "Token address must not be null"
+        );
+        require(_sendTokenAmount > 0 && _receiveTokenAmount > 0, "Token amount must be greater than 0");
+
+        _sendToken.safeTransferFrom(msg.sender, address(this), _sendTokenAmount);
+        bytes32 proposalId = votingMachine.propose(2, voteParamsHash, msg.sender, address(avatar));
```
orenyodfat, line 151: proposalId can be defined as return value at https://github.com/daostack/arc/pull/774/files#diff-0568445d702e1cdbb17b4e3cdc934e5dR142

contracts/schemes/TokenTrade.sol
```diff
@@ -0,0 +1,179 @@
+pragma solidity ^0.6.10;
+// SPDX-License-Identifier: GPL-3.0
+
+import "../votingMachines/VotingMachineCallbacks.sol";
+
+
+/**
+ * @title A scheme for join in a dao.
+ * - A member can be proposed to join in by sending a min amount of fee.
```
orenyodfat, line 9: this comment should be updated

contracts/schemes/TokenTrade.sol
```diff
@@ -0,0 +1,171 @@
...
+
+        emit ProposalExecuted(address(avatar), _proposalId, _decision);
+        return true;
+    }
+
+
+    function execute(bytes32 _proposalId) public {
+        Proposal storage proposal = proposals[_proposalId];
+        require(address(_sendToken) != address(0), "must be a live proposal");
+        require(proposal.decided, "must be a decided proposal");
```
orenyodfat, line 94: does this condition is not enough ? why need the previous one ?

PR #775, opened 2020-07-23
Add reputation token trade scheme

ben-kaufman:


contracts/schemes/ReputationTokenTrade.sol
```diff
@@ -0,0 +1,158 @@
...
+
+            emit TokenTradeProposalExecuted(
+                address(avatar),
+                _proposalId,
+                proposal.beneficiary,
+                proposal.sendToken,
+                proposal.sendTokenAmount,
+                proposal.reputationAmount
+            );
+            delete proposals[_proposalId];
```
orenyodfat, line 106: can delete proposal be moved out of the if and make it one time

contracts/schemes/ReputationTokenTrade.sol
```diff
@@ -0,0 +1,156 @@
...
+    * @param _proposalId the ID of the voting in the voting machine
+    * @param _decision a parameter of the voting result, 1 yes and 2 is no.
+    * @return bool success
+    */
+    function executeProposal(bytes32 _proposalId, int256 _decision)
+    external
+    onlyVotingMachine(_proposalId)
+    override
+    returns(bool) {
+        Proposal memory proposal = proposals[_proposalId];
```
orenyodfat, line 77: unused var ?

orenyodfat: also at https://github.com/daostack/arc/blob/master-2/contracts/schemes/TokenTrade.sol#L80

contracts/schemes/ReputationTokenTrade.sol
```diff
@@ -0,0 +1,155 @@
...
+        }
+        proposals[_proposalId].decided = true;
+
+        emit ProposalExecuted(address(avatar), _proposalId, _decision);
+        return true;
+    }
+
+
+    function execute(bytes32 _proposalId) public {
+        Proposal storage proposal = proposals[_proposalId];
```
orenyodfat, line 88: why this is `storage` ?

contracts/schemes/ReputationTokenTrade.sol
```diff
@@ -0,0 +1,155 @@
...
+                _proposalId,
+                proposal.beneficiary,
+                proposal.sendToken,
+                proposal.sendTokenAmount,
+                proposal.reputationAmount
+            );
+        } else {
+            proposal.sendToken.safeTransfer(address(proposal.beneficiary), proposal.sendTokenAmount);
+        }
+        delete proposals[_proposalId];
```
orenyodfat, line 108: does` delete proposal` will work ?

ben-kaufman: I changed it to `memory` so I don't think so.

PR #780, opened 2020-08-02
Only join

orenyodfat:


contracts/schemes/Join.sol
```diff
@@ -127,13 +104,12 @@ contract JoinAndQuit is
     returns(bool) {
         Proposal memory proposal = proposals[_proposalId];
         require(proposal.proposedMember != address(0), "not a valid proposal");
-        require(fundings[proposal.proposedMember].state == MemeberState.Candidate, "proposal already been executed");
+        require(membersState[proposal.proposedMember] == MemeberState.Candidate, "proposal already been executed");
```
jellegerbrandy, line 107: if memberState is "candidate", that means that the proposal is already executed? Maybe change the name of the state to something more meaningful then (like "executed"?)

contracts/schemes/Join.sol
```diff
@@ -127,13 +104,12 @@ contract JoinAndQuit is
     returns(bool) {
         Proposal memory proposal = proposals[_proposalId];
         require(proposal.proposedMember != address(0), "not a valid proposal");
```
jellegerbrandy, line 106: maybe give a more specific error message here?

contracts/schemes/Join.sol
```diff
@@ -59,22 +46,15 @@ contract JoinAndQuit is
...
 
-    struct MemberFund {
-        MemeberState state;
-        bool rageQuit;
-        uint256 funding;
-    }
-
     mapping(bytes32=>Proposal) public proposals;
-    mapping(address=>MemberFund) public fundings;
+    mapping(address=>MemeberState) public membersState;
```
jellegerbrandy, line 50: "MemberState" not "MemeberState"

contracts/schemes/Join.sol
```diff
@@ -212,11 +188,11 @@ contract JoinAndQuit is
     function redeemReputation(bytes32 _proposalId) public returns(uint256 reputation) {
         Proposal memory proposal = proposals[_proposalId];
         require(proposal.proposedMember != address(0), "no member to redeem");
-        require(!fundings[proposal.proposedMember].rageQuit, "member already rageQuit");
-        require(fundings[proposal.proposedMember].state == MemeberState.Accepted, "member not accepeted");
+        require(membersState[proposal.proposedMember] == MemeberState.Accepted, "member not accepeted");
```
jellegerbrandy, line 191: accepted, not accepted

contracts/schemes/Join.sol
```diff
@@ -174,11 +150,11 @@ contract JoinAndQuit is
     returns(bytes32)
     {
         address proposer = msg.sender;
-        require(fundings[proposer].state != MemeberState.Candidate, "already a candidate");
-        require(fundings[proposer].state != MemeberState.Accepted, "accepted and not redeemed yet");
+        require(membersState[proposer] != MemeberState.Candidate, "already a candidate");
+        require(membersState[proposer] != MemeberState.Accepted, "accepted and not redeemed yet");
         require(avatar.nativeReputation().balanceOf(proposer) == 0, "already a member");
         require(_feeAmount >= minFeeToJoin, "_feeAmount should be >= then the minFeeToJoin");
```
jellegerbrandy, line 156: "than" not "then"

contracts/schemes/Join.sol
```diff
@@ -174,11 +150,11 @@ contract JoinAndQuit is
     returns(bytes32)
     {
         address proposer = msg.sender;
-        require(fundings[proposer].state != MemeberState.Candidate, "already a candidate");
-        require(fundings[proposer].state != MemeberState.Accepted, "accepted and not redeemed yet");
+        require(membersState[proposer] != MemeberState.Candidate, "already a candidate");
+        require(membersState[proposer] != MemeberState.Accepted, "accepted and not redeemed yet");
```
jellegerbrandy, line 154: maybe add a subject to the error message so we know what it is about?

contracts/schemes/Join.sol
```diff
@@ -174,11 +150,11 @@ contract JoinAndQuit is
     returns(bytes32)
     {
         address proposer = msg.sender;
-        require(fundings[proposer].state != MemeberState.Candidate, "already a candidate");
-        require(fundings[proposer].state != MemeberState.Accepted, "accepted and not redeemed yet");
+        require(membersState[proposer] != MemeberState.Candidate, "already a candidate");
```
jellegerbrandy, line 153: maybe add a subject to the error message so we know what it is about? ("proposer is already a candidate")

PR #784, opened 2020-08-31
Added GenericSchemeMultiCall

nicoelzer:
- Added GenericSchemeMultiCall

### DIFF
--- a/contracts/schemes/GenericSchemeMultiCall.sol
+++ b/contracts/schemes/GenericSchemeMultiCall.sol
@@ -6,11 +6,11 @@ import "@daostack/infra/contracts/votingMachines/ProposalExecuteInterface.sol";
 import "../votingMachines/VotingMachineCallbacks.sol";
 
 /**
- * @title GenericScheme.
+ * @title GenericSchemeMultiCall.
  * @dev  A scheme for proposing and executing calls to multiple arbitrary function
- * on whitelisted contract on behalf of the organization avatar.
+ * on one or multiple contracts on behalf of the organization avatar.
  */
-contract GenericSchemeMulticall is VotingMachineCallbacks, ProposalExecuteInterface {
+contract GenericSchemeMultiCall is VotingMachineCallbacks, ProposalExecuteInterface {
     event NewMultiCallProposal(
         address indexed _avatar,
         bytes32 indexed _proposalId,
--- a/contracts/schemes/GenericSchemeMultiCall.sol
+++ b/contracts/schemes/GenericSchemeMultiCall.sol
@@ -140,7 +140,6 @@ contract GenericSchemeMultiCall is VotingMachineCallbacks, ProposalExecuteInterf
            );
            controller.externalTokenApproval(extToken,spender,valueToSpend,avatar);
          } else {
-           require(contractWhitelist[proposal.contractsToCall[i]], "contractToCall is not whitelisted");
            (success, genericCallReturnValue) =
            controller.genericCall(proposal.contractsToCall[i], proposal.callData[i], avatar, proposal.value[i]);
          }
@@ -171,6 +170,12 @@ contract GenericSchemeMultiCall is VotingMachineCallbacks, ProposalExecuteInterf
             (_contractsToCall.length == _callData.length) && (_contractsToCall.length == _value.length),
             "Wrong length of _contractsToCall, _callData or _value arrays"
         );
+        for (uint i = 0; i < _contractsToCall.length; i ++) {
+            require(
+                contractWhitelist[_contractsToCall[i]] || _contractsToCall[i] == avatar.owner(),
+                 "contractToCall is not whitelisted"
+            );
+        }
         proposalId = votingMachine.propose(2, voteParams, msg.sender, address(avatar));
 
         proposals[proposalId] = MultiCallProposal({
--- a/contracts/schemes/GenericSchemeMultiCall.sol
+++ b/contracts/schemes/GenericSchemeMultiCall.sol
@@ -1,0 +1,167 @@
+pragma solidity 0.5.17;
+pragma experimental ABIEncoderV2;
+
+import "@daostack/infra/contracts/votingMachines/IntVoteInterface.sol";
+import "@daostack/infra/contracts/votingMachines/ProposalExecuteInterface.sol";
+import "../votingMachines/VotingMachineCallbacks.sol";
+
+/**
+ * @title GenericScheme.
+ * @dev  A scheme for proposing and executing calls to multiple arbitrary function
+ * on whitelisted contract on behalf of the organization avatar.
+ */
+contract GenericSchemeMulticall is VotingMachineCallbacks, ProposalExecuteInterface {
+    event NewMultiCallProposal(
+        address indexed _avatar,
+        bytes32 indexed _proposalId,
+        bytes[]   _callData,
+        uint[] _value,
+        string  _descriptionHash,
+        address[] _contractToCall
+    );
+
+    event ProposalExecuted(
+        address indexed _avatar,
+        bytes32 indexed _proposalId
+    );
+
+    event ProposalCallExecuted(
+        address indexed _avatar,
+        bytes32 indexed _proposalId,
+        address _contractToCall,
+        bytes _callDataReturnValue
+    );
+
+    event ProposalExecutedByVotingMachine(
+        address indexed _avatar,
+        bytes32 indexed _proposalId,
+        int256 _param
+    );
+
+    event ProposalDeleted(address indexed _avatar, bytes32 indexed _proposalId);
+
+    // Details of a voting proposal:
+    struct MultiCallProposal {
+        address[] contractToCall;
+        bytes[] callData;
+        uint256[] value;
+        bool exist;
+        bool passed;
+    }
+
+    mapping(bytes32=>MultiCallProposal) public organizationProposals;
+
+    IntVoteInterface public votingMachine;
+    bytes32 public voteParams;
+    Avatar public avatar;
+
+    /**
+     * @dev initialize
+     * @param _avatar the avatar to mint reputation from
+     * @param _votingMachine the voting machines address to
+     * @param _voteParams voting machine parameters.
+     */
+    function initialize(
+        Avatar _avatar,
+        IntVoteInterface _votingMachine,
+        bytes32 _voteParams
+    )
+    external
+    {
+        require(avatar == Avatar(0), "can be called only one time");
+        require(_avatar != Avatar(0), "avatar cannot be zero");
+        avatar = _avatar;
+        votingMachine = _votingMachine;
+        voteParams = _voteParams;
+    }
+
+    /**
+    * @dev execution of proposals, can only be called by the voting machine in which the vote is held.
+    * @param _proposalId the ID of the voting in the voting machine
+    * @param _decision a parameter of the voting result, 1 yes and 2 is no.
+    * @return bool success
+    */
+    function executeProposal(bytes32 _proposalId, int256 _decision)
+    external
+    onlyVotingMachine(_proposalId)
+    returns(bool) {
+        MultiCallProposal storage proposal = organizationProposals[_proposalId];
+        require(proposal.exist, "must be a live proposal");
+        require(proposal.passed == false, "cannot execute twice");
+
+        if (_decision == 1) {
+            proposal.passed = true;
+            execute(_proposalId);
+        } else {
+            delete organizationProposals[_proposalId];
+            emit ProposalDeleted(address(avatar), _proposalId);
+        }
+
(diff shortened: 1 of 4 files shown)


contracts/schemes/GenericSchemeMultiCall.sol
```diff
@@ -0,0 +1,167 @@
...
+ * @title GenericScheme.
+ * @dev  A scheme for proposing and executing calls to multiple arbitrary function
+ * on whitelisted contract on behalf of the organization avatar.
+ */
+contract GenericSchemeMulticall is VotingMachineCallbacks, ProposalExecuteInterface {
+    event NewMultiCallProposal(
+        address indexed _avatar,
+        bytes32 indexed _proposalId,
+        bytes[]   _callData,
+        uint[] _value,
```
orenyodfat, line 18: uint256[]

contracts/schemes/GenericSchemeMultiCall.sol
```diff
@@ -0,0 +1,167 @@
...
+    // Details of a voting proposal:
+    struct MultiCallProposal {
+        address[] contractToCall;
+        bytes[] callData;
+        uint256[] value;
+        bool exist;
+        bool passed;
+    }
+
+    mapping(bytes32=>MultiCallProposal) public organizationProposals;
```
orenyodfat, line 52: rename organizationProposals -> proposals

contracts/schemes/GenericSchemeMultiCall.sol
```diff
@@ -0,0 +1,167 @@
...
+    * @dev propose to call on behalf of the _avatar
+    *      The function trigger NewCallProposal event
+    * @param _contractToCall the contract to be called 
+    * @param _callData - The abi encode data for the call
+    * @param _value value(ETH) to transfer with the call
+    * @param _descriptionHash proposal description hash
+    * @return an id which represents the proposal
+    */
+
+    function proposeCall(address[] memory _contractToCall, bytes[] memory _callData, uint256[] memory _value, string memory _descriptionHash)
```
orenyodfat, line 140: `proposeCalls` ?

contracts/schemes/GenericSchemeMultiCall.sol
```diff
@@ -0,0 +1,167 @@
...
+        delete organizationProposals[_proposalId];
+        emit ProposalDeleted(address(avatar), _proposalId);
+        emit ProposalExecuted(address(avatar), _proposalId);
+
+    }
+
+    /**
+    * @dev propose to call on behalf of the _avatar
+    *      The function trigger NewCallProposal event
+    * @param _contractToCall the contract to be called
```
orenyodfat, line 133: description is wrong..

contracts/schemes/GenericSchemeMultiCall.sol
```diff
@@ -0,0 +1,167 @@
...
+        require(proposal.passed, "proposal must passed by voting machine");
+        proposal.exist = false;
+        bytes memory genericCallReturnValue;
+        bool success;
+        Controller controller = Controller(avatar.owner());
+
+        for (uint i = 0; i < proposal.contractToCall.length; i ++) {
+            (success, genericCallReturnValue) =
+            controller.genericCall(proposal.contractToCall[i], proposal.callData[i], avatar, proposal.value[i]);
+            require(success, "Proposal call failed");
```
orenyodfat, line 120: so if one of the calls fails all their is no execution .. maybe add a comment re that

contracts/schemes/GenericSchemeMultiCall.sol
```diff
@@ -0,0 +1,167 @@
...
+    /**
+     * @dev initialize
+     * @param _avatar the avatar to mint reputation from
+     * @param _votingMachine the voting machines address to
+     * @param _voteParams voting machine parameters.
+     */
+    function initialize(
+        Avatar _avatar,
+        IntVoteInterface _votingMachine,
+        bytes32 _voteParams
```
orenyodfat, line 67: how about adding here a white listed targets contracts? for security reasons.

contracts/schemes/GenericSchemeMultiCall.sol
orenyodfat, line 140: _contractsToCall

contracts/schemes/GenericSchemeMultiCall.sol
```diff
@@ -0,0 +1,167 @@
...
+    * @param _contractToCall the contract to be called 
+    * @param _callData - The abi encode data for the call
+    * @param _value value(ETH) to transfer with the call
+    * @param _descriptionHash proposal description hash
+    * @return an id which represents the proposal
+    */
+
+    function proposeCall(address[] memory _contractToCall, bytes[] memory _callData, uint256[] memory _value, string memory _descriptionHash)
+    public
+    returns(bytes32)
```
orenyodfat, line 142: `returns(bytes32 proposalId)`

contracts/schemes/GenericSchemeMultiCall.sol
```diff
@@ -0,0 +1,167 @@
...
+            exist: true,
+            passed: false
+        });
+        proposalsInfo[address(votingMachine)][proposalId] = ProposalInfo({
+            blockNumber:block.number,
+            avatar:avatar
+        });
+
+        emit NewMultiCallProposal(address(avatar), proposalId, _callData, _value, _descriptionHash, _contractToCall);
+        return proposalId;
```
orenyodfat, line 163: can be removed

contracts/schemes/GenericSchemeMultiCall.sol
```diff
@@ -0,0 +1,167 @@
+pragma solidity 0.5.17;
+pragma experimental ABIEncoderV2;
```
orenyodfat, line 2: why this is needed ?

orenyodfat (changes requested):

contracts/schemes/GenericSchemeMultiCall.sol
nicoelzer: @orenyodfat bytes[]  _callData is only supported with ABIEncoderV2 in 0.5

contracts/schemes/GenericSchemeMultiCall.sol
nicoelzer: Why removing it? We would still want to emit the event and return proposalId after creating the Proposal

contracts/schemes/GenericSchemeMultiCall.sol
orenyodfat: ok. the other alternative is to pass in bytes _callData which is concatenation of call data of all calls .
and provide additional lens array to parse it.
something like https://github.com/daostack/arc/blob/master-2/contracts/utils/DAOFactory.sol#L235 
this will eliminate the need to use the `experimental` pragma which is not experimental.

contracts/schemes/GenericSchemeMultiCall.sol
nicoelzer: @orenyodfat as we would need to call allowance() on some tokens we don't know at initialization we won't be able to whitelist specific contract addresses. Thinking about a possible solution for restriction right now.

orenyodfat: losing a bit from generality , one option is to add explicit none restricted `approval` token calls as part of a proposal .
using https://github.com/daostack/arc/blob/master/contracts/controller/Controller.sol#L460

nicoelzer: We could still keep the generality if we would include an encoded call within the proposalCalls function that would execute externalTokenApproval() on the controller before sending funds. In that way we could still use a contract whitelist as we would always use the controller (known address) for approvals – correct?

orenyodfat: > @orenyodfat we could still keep the generality if we would include an encoded call within the proposalCalls function that would execute externalTokenApproval() on the controller before sending funds. In that way we could still use a contract whitelist as we would always use the controller (known address) for approvals – correct?

no. because the avatar is not allowed to call controller . it is not registered scheme

orenyodfat: > @orenyodfat we could still keep the generality if we would include an encoded call within the proposalCalls function that would execute externalTokenApproval() on the controller before sending funds. In that way we could still use a contract whitelist as we would always use the controller (known address) for approvals – correct?

no . because the avatar is not allowed to call controller .(it is not a register scheme)

contracts/schemes/GenericSchemeMultiCall.sol
nicoelzer: @orenyodfat the DAOfactory is also using ABIEncoderV2, so I guess that shouldn't be an issue

orenyodfat: DAOFactory use solidity 6.12

contracts/schemes/GenericSchemeMultiCall.sol
```diff
@@ -0,0 +1,192 @@
...
+            uint256 valueToSpend
+           ) =
+           /* solhint-disable */
+           abi.decode(
+             proposal.callData[i],
+             (IERC20, address, uint256)
+           );
+           controller.externalTokenApproval(extToken,spender,valueToSpend,avatar);
+         } else {
+           require(contractWhitelist[proposal.contractsToCall[i]], "contractToCall is not whitelisted");
```
orenyodfat, line 143: this should be check at the proposal function

test/genericscheme.js
```diff
@@ -232,7 +232,7 @@ contract('GenericScheme', function(accounts) {
         });
       });
 
-      it("Wallet - execute proposeVote -positive decision - check action - with GenesisProtocol", async function() {
+      it.only("Wallet - execute proposeVote -positive decision - check action - with GenesisProtocol", async function() {
```
orenyodfat, line 235: is this related to this pr ?

contracts/schemes/GenericSchemeMultiCall.sol
```diff
@@ -0,0 +1,192 @@
...
+    * @param _value value(ETH) to transfer with the calls
+    * @param _descriptionHash proposal description hash
+    * @return an id which represents the proposal
+    */
+
+    function proposeCalls(address[] memory _contractsToCall, bytes[] memory _callData, uint256[] memory _value, string memory _descriptionHash)
+    public
+    returns(bytes32 proposalId)
+    {
+        require(
```
orenyodfat, line 170: is there a test covering this one ?

contracts/schemes/GenericSchemeMultiCall.sol
orenyodfat: any test cover proposing none white list contracts ?

contracts/schemes/GenericSchemeMultiCall.sol
```diff
@@ -0,0 +1,192 @@
...
+           (IERC20 extToken,
+            address spender,
+            uint256 valueToSpend
+           ) =
+           /* solhint-disable */
+           abi.decode(
+             proposal.callData[i],
+             (IERC20, address, uint256)
+           );
+           controller.externalTokenApproval(extToken,spender,valueToSpend,avatar);
```
orenyodfat, line 141: any test cover this case ?

contracts/schemes/GenericSchemeMultiCall.sol
nicoelzer: > any test cover proposing none white list contracts ?

it("execute proposeVote -positive decision - not whitelisted contract")

contracts/schemes/GenericSchemeMultiCall.sol
nicoelzer: > is there a test covering this one ?

@orenyodfat it("proposeCall log - with invalid array - reverts")

contracts/schemes/GenericSchemeMultiCall.sol
nicoelzer: it("execute proposeVote with multiple calls with votingMachine -positive decision")

contracts/schemes/GenericSchemeMultiCall.sol
AugustoL: @orenyodfat the avatar wont call the controller it would called an ERC20 contract and execute the approve function.

AugustoL: But taht will require the ERC20 token to be listed in the whitelisted contracts, right? this way the controller can call the approve function in any token contract, right? or am I missing something?

orenyodfat: @nicoelzer  does any how the plan is to allow the avatar to transfer any erc20 token its own?. this might be the same as approval ... if this is the case we might do not need this special approval staff.
If the case is that only approval is needed for any erc20 so we can avoid white listing the erc20 and have the solution as it is now .

PR #786, opened 2020-09-03
Added GenericSchemeMultiCall

nicoelzer:
- Added GenericSchemeMultiCall


contracts/schemes/GenericSchemeMultiCall.sol
```diff
@@ -0,0 +1,197 @@
...
+        MultiCallProposal storage proposal = proposals[_proposalId];
+        require(proposal.exist, "must be a live proposal");
+        require(proposal.passed, "proposal must passed by voting machine");
+        proposal.exist = false;
+        bytes memory genericCallReturnValue;
+        bool success;
+        Controller controller = Controller(avatar.owner());
+
+        for (uint i = 0; i < proposal.contractsToCall.length; i ++) {
+          if (proposal.contractsToCall[i] == address(controller)) {
```
AugustoL, line 130: why is this condition check needed? The controller can just called the token approval to the avatar with the encoded data, right?

contracts/schemes/GenericSchemeMultiCall.sol
```diff
@@ -0,0 +1,197 @@
...
+    public
+    returns(bytes32 proposalId)
+    {
+        require(
+            (_contractsToCall.length == _callData.length) && (_contractsToCall.length == _value.length),
+            "Wrong length of _contractsToCall, _callData or _value arrays"
+        );
+        for (uint i = 0; i < _contractsToCall.length; i ++) {
+            require(
+                contractWhitelist[_contractsToCall[i]] || _contractsToCall[i] == avatar.owner(),
```
AugustoL, line 175: Why checking against the avatar owner that is going to be the controller, do we want this scheme to execute controller functions besides genericCalls?

contracts/schemes/GenericSchemeMultiCall.sol
nicoelzer: see this convo: https://github.com/daostack/arc/pull/784#discussion_r480865631

contracts/schemes/GenericSchemeMultiCall.sol
```diff
@@ -0,0 +1,197 @@
...
+            uint256 valueToSpend
+           ) =
+           /* solhint-disable */
+           abi.decode(
+             proposal.callData[i],
+             (IERC20, address, uint256)
+           );
+           (success) = controller.externalTokenApproval(extToken,spender,valueToSpend,avatar);
+         } else {
+           (success, genericCallReturnValue) =
```
AugustoL, line 143: if the first 4 bytes of the `genericCallReturnValue` are `0x08c379a0` that would mean that the call failed because the error message was throw but `success` can equals true, shouldnt we take that in count as call failed?

AugustoL: reference: https://github.com/AugustoL/dxdao-contracts/blob/dxdao-contracts/contracts/schemes/WalletScheme.sol#L103

contracts/schemes/GenericSchemeMultiCall.sol
AugustoL: Just commented it, I think there is a misunderstanding there because the avatar will call the ERC20 token, no the controller. 

But if you dont check that you have to list every ERC20 token that you want to call, I guess calling the external token transfer in the controller allows to call any token, it is like a "whitelisted" function

contracts/schemes/GenericSchemeMultiCall.sol
nicoelzer: Yes, we would use the externalTokenApproval() on the controller to set the token allowance

contracts/schemes/GenericSchemeMultiCall.sol
orenyodfat: @nicoelzer does any how the plan is to allow the avatar to transfer any erc20 token its own?. this might be the same as approval ... if this is the case we might do not need this special approval staff. as @AugustoL mention. (though will need to whitelist that somehow).
If the case is that only approval is needed for any erc20 so we can avoid white listing the erc20 and have the solution as it is now .

contracts/schemes/GenericSchemeMultiCall.sol
orenyodfat: Is there any documentation re this issue ?  I found this [old](https://github.com/ethereum/go-ethereum/issues/19027) one. 
Also , no reference for that in [OZ call](https://github.com/OpenZeppelin/openzeppelin-contracts/blob/c801c8d2bbaeeab413fc792ae0d44f720fb81bc9/contracts/utils/Address.sol#L119) implementation.
@AugustoL is that reproducible ?

contracts/schemes/GenericSchemeMultiCall.sol
```diff
@@ -0,0 +1,197 @@
...
+    function proposeCalls(address[] memory _contractsToCall, bytes[] memory _callData, uint256[] memory _value, string memory _descriptionHash)
+    public
+    returns(bytes32 proposalId)
+    {
+        require(
+            (_contractsToCall.length == _callData.length) && (_contractsToCall.length == _value.length),
+            "Wrong length of _contractsToCall, _callData or _value arrays"
+        );
+        for (uint i = 0; i < _contractsToCall.length; i ++) {
+            require(
```
orenyodfat, line 174: add a require that the spender is whitelisted for the case of _contractsToCall is controller.

contracts/schemes/GenericSchemeMultiCall.sol
```diff
@@ -0,0 +1,197 @@
...
+                (IERC20 extToken,
+                address spender,
+                uint256 valueToSpend
+                ) =
+                /* solhint-disable */
+                abi.decode(
+                    proposal.callData[i],
+                    (IERC20, address, uint256)
+                );
+           require(contractWhitelist[spender], "spender contract not whitelisted");
```
orenyodfat, line 140: It might be better to catch that at the propose function.

contracts/schemes/GenericSchemeMultiCall.sol
```diff
@@ -0,0 +1,197 @@
...
+    public
+    returns(bytes32 proposalId)
+    {
+        require(
+            (_contractsToCall.length == _callData.length) && (_contractsToCall.length == _value.length),
+            "Wrong length of _contractsToCall, _callData or _value arrays"
+        );
+        for (uint i = 0; i < _contractsToCall.length; i ++) {
+            require(
+                contractWhitelist[_contractsToCall[i]] || _contractsToCall[i] == avatar.owner(),
```
orenyodfat, line 174: this check can be simplify by adding the controller to contractWhitelist at initilize function .

contracts/schemes/GenericSchemeMultiCall.sol
```diff
@@ -0,0 +1,197 @@
...
+        uint256[] value;
+        bool exist;
+        bool passed;
+    }
+
+    mapping(bytes32=>MultiCallProposal) public proposals;
+
+    IntVoteInterface public votingMachine;
+    bytes32 public voteParams;
+    mapping(address=>bool) internal contractWhitelist;
```
orenyodfat, line 28: how about make it public ?

nicoelzer: @orenyodfat whitelistedContracts is public, having both public would probably confuse

.env
```diff
@@ -0,0 +1 @@
+ETHERSCAN_API_KEY=6B66EZA1S6B26FCMF4I4DC1FVUZTZKIK2S
```
orenyodfat, line 1: is that a secret ?

.gitignore
```diff
@@ -8,3 +8,4 @@ node_modules/
 .vscode/
 site/
 yarn*
+.secret
```
orenyodfat, line 11: what is .secret ? can you remove it ?

contracts/schemes/GenericSchemeMultiCall.sol
AugustoL: @orenyodfat I think it is enough the way it is done by checking only the `bool success`, the signature of the data returned will be `0x08c379a0` in case there is a revert reason/message.

contracts/schemes/GenericSchemeMultiCall.sol
AugustoL: cool, maybe we can just add the controller to the contractWhitelist by default.

.gitignore
```diff
@@ -7,4 +7,4 @@ node_modules/
 *.tgz
 .vscode/
 site/
-yarn*
+yarn*
```
orenyodfat, line 10: please remove this new line..

contracts/schemes/GenericSchemeMultiCall.sol
```diff
@@ -0,0 +1,213 @@
...
+    * @param _proposalId the ID of the voting in the voting machine
+    */
+    function execute(bytes32 _proposalId) public {
+        MultiCallProposal storage proposal = proposals[_proposalId];
+        require(proposal.exist, "must be a live proposal");
+        require(proposal.passed, "proposal must passed by voting machine");
+        proposal.exist = false;
+        bytes memory genericCallReturnValue;
+        bool success;
+        Controller controller = Controller(avatar.owner());
```
orenyodfat, line 132: redundant call . 
you can use : 
`Controller controller = Controller(whitelistedContracts[0]);`

contracts/schemes/GenericSchemeMultiCall.sol
```diff
@@ -0,0 +1,213 @@
...
+
+    function proposeCalls(address[] memory _contractsToCall, bytes[] memory _callData, uint256[] memory _value, string memory _descriptionHash)
+    public
+    returns(bytes32 proposalId)
+    {
+        require(
+            (_contractsToCall.length == _callData.length) && (_contractsToCall.length == _value.length),
+            "Wrong length of _contractsToCall, _callData or _value arrays"
+        );
+        Controller controller = Controller(avatar.owner());
```
orenyodfat, line 177: edundant call . 
you can use : 
`Controller controller = Controller(whitelistedContracts[0]);`

contracts/schemes/GenericSchemeMultiCall.sol
```diff
@@ -0,0 +1,213 @@
...
+        proposalsInfo[address(votingMachine)][proposalId] = ProposalInfo({
+            blockNumber:block.number,
+            avatar:avatar
+        });
+
+        emit NewMultiCallProposal(address(avatar), proposalId, _callData, _value, _descriptionHash, _contractsToCall);
+
+    }
+
+}
```
orenyodfat, line 213: remove this

.gitignore
```diff
@@ -6,5 +6,4 @@ node_modules/
 .node-xmlhttprequest-sync*
 *.tgz
 .vscode/
-site/
-yarn*
+site/
```
orenyodfat, line 9: this .gitignore file should not be changed at all

contracts/schemes/GenericSchemeMultiCall.sol
```diff
@@ -0,0 +1,213 @@
...
+        for (uint i = 0; i < _contractsToCall.length; i ++) {
+            require(
+                contractWhitelist[_contractsToCall[i]], "contractToCall is not whitelisted"
+            );
+            if (_contractsToCall[i] == address(controller)) {
+                (IERC20 extToken,
+                address spender,
+                uint256 valueToSpend
+                ) =
+                /* solhint-disable */
```
orenyodfat, line 187: please add more specific solvent-diable-...? 
or re enable that after the problematic lines. I think at line 192

contracts/schemes/GenericSchemeMultiCall.sol
```diff
@@ -0,0 +1,213 @@
...
+                abi.decode(
+                    proposal.callData[i],
+                    (IERC20, address, uint256)
+                );
+           (success) = controller.externalTokenApproval(extToken,spender,valueToSpend,avatar);
```
orenyodfat, line 145: /* solhint-enable */

contracts/schemes/GenericSchemeMultiCall.sol
```diff
@@ -0,0 +1,212 @@
...
+    * @dev propose to call one or multiple contracts on behalf of the _avatar
+    *      The function trigger NewMultiCallProposal event
+    * @param _contractsToCall the contracts to be called 
+    * @param _callData - The abi encode data for the calls
+    * @param _value value(ETH) to transfer with the calls
+    * @param _descriptionHash proposal description hash
+    * @return an id which represents the proposal
+    */
+
+    function proposeCalls(address[] memory _contractsToCall, bytes[] memory _callData, uint256[] memory _value, string memory _descriptionHash)
```
orenyodfat, line 169: _callData -> _callsData
_value -> values

contracts/schemes/GenericSchemeMultiCall.sol
```diff
@@ -0,0 +1,212 @@
...
+    IntVoteInterface public votingMachine;
+    bytes32 public voteParams;
+    mapping(address=>bool) internal contractWhitelist;
+    address[] public whitelistedContracts;
+    Avatar public avatar;
+
+    event NewMultiCallProposal(
+        address indexed _avatar,
+        bytes32 indexed _proposalId,
+        bytes[]   _callData,
```
orenyodfat, line 35: naming : 
_callData -> _callsData
_value -> values

.gitignore
orenyodfat, line 10: why this file changed at all ?

package.json
```diff
@@ -76,6 +77,8 @@
   "homepage": "https://daostack.io",
   "dependencies": {
     "@daostack/infra": "0.0.1-rc.19",
+    "@truffle/hdwallet-provider": "^1.0.43",
```
orenyodfat, line 80: why this is needed ?

package.json
```diff
@@ -76,6 +77,8 @@
...
+    "dotenv": "^8.2.0",
```
orenyodfat, line 81: why this is needed

orenyodfat (approved):

PR #788, opened 2020-09-16
Updated GenericSchemeMultiCall revert on failed calls

nicoelzer:
- Updated GenericSchemeMultiCall revert on failed calls
- Updated GenericSchemeMultiCall Bool Literals
- Updated GenericSchmeMultiCall events

### DIFF
--- a/contracts/schemes/GenericSchemeMultiCall.sol
+++ b/contracts/schemes/GenericSchemeMultiCall.sol
@@ -152,9 +152,10 @@ contract GenericSchemeMultiCall is VotingMachineCallbacks, ProposalExecuteInterf
             } else {
                 (success, genericCallReturnValue) =
                 controller.genericCall(proposal.contractsToCall[i], callData, avatar, proposal.values[i]);
-                /* Whole transaction will be reverted if at least one call fails*/
-                require(success, "Proposal call failed");
             }
+
+            /* Whole transaction will be reverted if at least one call fails*/
+            require(success, "Proposal call failed");
             startIndex = startIndex.add(proposal.callsDataLens[i]);
             emit ProposalCallExecuted(
                 address(avatar),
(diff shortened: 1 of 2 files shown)


contracts/schemes/GenericSchemeMultiCall.sol
```diff
@@ -152,6 +152,8 @@ contract GenericSchemeMultiCall is VotingMachineCallbacks, ProposalExecuteInterf
             } else {
                 (success, genericCallReturnValue) =
                 controller.genericCall(proposal.contractsToCall[i], callData, avatar, proposal.values[i]);
+                /* Whole transaction will be reverted if at least one call fails*/
+                require(success, "Proposal call failed");
```
orenyodfat, line 156: please move that out of the` }` so it will include the `externalTokenApproval` as well .

nicoelzer: @orenyodfat just updated.

contracts/schemes/GenericSchemeMultiCall.sol
```diff
@@ -153,6 +153,9 @@ contract GenericSchemeMultiCall is VotingMachineCallbacks, ProposalExecuteInterf
                 (success, genericCallReturnValue) =
                 controller.genericCall(proposal.contractsToCall[i], callData, avatar, proposal.values[i]);
             }
+
+            /* Whole transaction will be reverted if at least one call fails*/
+            require(success, "Proposal call failed");
```
orenyodfat, line 158: please remove the automatic call to `execute`  at https://github.com/daostack/arc/blob/b80fe92ff7dd9eb069c7770431c5dfb041d3e9e1/contracts/schemes/GenericSchemeMultiCall.sol#L116 .
this is not good due to :
1. last voter(executer) does not need to pay for whole execution gas.
2. if execution is reverted in one of the calls.. so the dao cannot decide (for the case of absolute majority)..
 - there suppose to be a differentiation between the dao decision and execution.

orenyodfat (approved):

ben-kaufman (approved):
Looks good to me

PR #795, opened 2020-10-08
GenericSchemeMultiCall  : Simplification. +  Schemeconstraint 

orenyodfat:


ben-kaufman (approved):

contracts/schemes/DxDaoSchemeConstraints.sol
```diff
@@ -0,0 +1,149 @@
...
+     * @param _contractsToCall the contracts to be called
+     * @param _callsData - The abi encode data for the calls
+     * @param _values value(ETH) to transfer with the calls
+     * @param _avatar avatar
+     * @return bool value true-allowed false not allowed
+     */
+    function isAllowedToPropose(
+        address[] calldata _contractsToCall,
+        bytes[] calldata _callsData,
+        uint256[] calldata,
```
nicoelzer, line 116: uint256[] calldata > uint256[] calldata _values?

contracts/schemes/DxDaoSchemeConstraints.sol
```diff
@@ -0,0 +1,149 @@
...
+                assembly {
+                    spender := mload(add(callData, 36))
+                }
+                require(contractsWhiteListMap[spender], "spender contract not whitelisted");
+            }
+        }
+        return true;
+    }
+
+    function observationIndex() public view returns (uint256) {
```
nicoelzer, line 144: @orenyodfat this could become vulnerable to overflows if the period is set to a super low level, I would propose to add a require in the initialize() to set the period to a min limit as of 7days.

contracts/schemes/DxDaoSchemeConstraints.sol
orenyodfat: yes, because the interface defined that function and there is no use for values at this function

PR #798, opened 2020-10-13
Added upgradable contraintLimits

nicoelzer:
- Added upgradable contraintLimits #798


contracts/schemes/DxDaoSchemeConstraints.sol
```diff
@@ -50,6 +54,51 @@ contract DxDaoSchemeConstraints is SchemeConstraints {
         contractsWhiteList = _contractsWhiteList;
     }
 
+    /*
+     * @dev updateContractWhitelist used to let the DAO update whitelisted contracts.
+     * @param _contractsAddresses - The contract that should be update
```
orenyodfat, line 59: contracts

contracts/schemes/DxDaoSchemeConstraints.sol
```diff
@@ -50,6 +54,51 @@ contract DxDaoSchemeConstraints is SchemeConstraints {
...
+        require(_contractsAddresses.length == _contractsWhitelisted.length,
+        "invalid length _periodLimitTokensAddresses");
+        for (uint i = 0; i < _contractsAddresses.length; i++) {
+            contractsWhiteListMap[_contractsAddresses[i]] = _contractsWhitelisted[i];
+        }
+    }
+
+    /*
+     * @dev updatePeriodLimitTokens lets the dao update limits to token limits.
+     * @param _tokensAddresses - The token that should be updated
```
orenyodfat, line 77: tokens

contracts/schemes/DxDaoSchemeConstraints.sol
```diff
@@ -50,6 +54,51 @@ contract DxDaoSchemeConstraints is SchemeConstraints {
...
+        for (uint i = 0; i < _tokensAddresses.length; i++) {
+            periodLimitToken[_tokensAddresses[i]] = _tokensPeriodLimits[i];
+        }
+    }
+
+    /*
+     * @dev updatePeriodLimitWei lets the dao update limits to ETH spending limit.
+     * @param _periodLimitWei - The new spending limit in WEI that should be set.
+     */
+    function updatePeriodLimitWei(uint256 _periodLimitWei) external {
```
orenyodfat, line 97: why not consolidate all these 3 functions to a single one ?

orenyodfat (changes requested):

orenyodfat:
so the idea is to have other genericscheme with different gov params which will  white list the schemeConstraint contract ?
other 2 alternative to do so are : 
1. add updateSchemeConstraint function to the genericSchemeMultiCall contract.
2. use the already exist plugin manager scheme to update the genericSchemeMultiCall as a whole. 

Please note that option 2 is already available and exist any how.. what is the rational/gain to add another way to do that ?

nicoelzer:
> so the idea is to have other genericscheme with different gov params which will white list the schemeConstraint contract ?
> other 2 alternative to do so are :
> 
> 1. add updateSchemeConstraint function to the genericSchemeMultiCall contract.
> 2. use the already exist plugin manager scheme to update the genericSchemeMultiCall as a whole.
> 
> Please note that option 2 is already available and exist any how.. what is the rational/gain to add another way to do that ?

Hi @orenyodfat,

exactly the idea would be to use another generic scheme with super safe governance parameters. A proposal should not be able to pass faster then 14 days. 

The reasons why a separate GenericScheme might be helpful for me are:
1. Differenciated, safer governance params then MultiCall & Plugin Manager
2. Being able to update the limits while the periodSpending are not lost. If we update the MultiCall as a whole we would also reset the spendingLimits of any token/eth.

orenyodfat:
> > so the idea is to have other genericscheme with different gov params which will white list the schemeConstraint contract ?
> > other 2 alternative to do so are :
> > 
> > 1. add updateSchemeConstraint function to the genericSchemeMultiCall contract.
> > 2. use the already exist plugin manager scheme to update the genericSchemeMultiCall as a whole.
> > 
> > Please note that option 2 is already available and exist any how.. what is the rational/gain to add another way to do that ?
> 
> Hi @orenyodfat,
> 
> exactly the idea would be to use another generic scheme with super safe governance parameters. A proposal should not be able to pass faster then 14 days.
> 
> The reasons why a separate GenericScheme might be helpful for me are:
> 
> 1. Differenciated, safer governance params then MultiCall & Plugin Manager
> 2. Being able to update the limits while the periodSpending are not lost. If we update the MultiCall as a whole we would also reset the spendingLimits of any token/eth.

this add another complexity to the system . another things to look on and monitor .

contracts/schemes/DxDaoSchemeConstraints.sol
```diff
@@ -83,6 +131,7 @@ contract DxDaoSchemeConstraints is SchemeConstraints {
                 callData[3] == APPROVE_SIGNATURE[3]) {
                 uint256 amount;
                 address contractToCall = _contractsToCall[i];
+                require(contractsWhiteListMap[contractToCall] == true, "tokenNotWhitelisted");
```
orenyodfat, line 134: this is redundant

contracts/schemes/DxDaoSchemeConstraints.sol
```diff
@@ -83,6 +131,7 @@ contract DxDaoSchemeConstraints is SchemeConstraints {
...
                 // solhint-disable-next-line no-inline-assembly
                 assembly {
                     amount := mload(add(callData, 68))
```
orenyodfat, line 137: if the whitelist is updated there is a need to re check the spender

orenyodfat (changes requested):
please emit an event on all updates functions

contracts/schemes/DxDaoSchemeConstraints.sol
```diff
@@ -3,10 +3,24 @@ pragma experimental ABIEncoderV2;
 
 import "./SchemeConstraints.sol";
 
-
 contract DxDaoSchemeConstraints is SchemeConstraints {
     using SafeMath for uint256;
 
+    event UpdatedContractsWhitelist(
+        address _contractAddress,
```
orenyodfat, line 10: how about passing the whole array .

contracts/schemes/DxDaoSchemeConstraints.sol
```diff
@@ -3,10 +3,24 @@ pragma experimental ABIEncoderV2;
...
 contract DxDaoSchemeConstraints is SchemeConstraints {
     using SafeMath for uint256;
 
+    event UpdatedContractsWhitelist(
+        address _contractAddress, 
+        bool _contractWhitelisted
+    );
+
+    event UpdatedPeriodLimitsTokens(
+        address _tokenAddress,
```
orenyodfat, line 15: just pass the whole array

contracts/schemes/DxDaoSchemeConstraints.sol
nicoelzer: Has been addresses in newest commit

contracts/schemes/DxDaoSchemeConstraints.sol
nicoelzer: Has been addresses in newest commit

orenyodfat (changes requested):
please base this pr vs master (or open a new one)

contracts/schemes/DxDaoSchemeConstraints.sol
```diff
@@ -82,7 +86,7 @@ contract DxDaoSchemeConstraints is SchemeConstraints {
         "invalid length _periodLimitTokensAddresses");
         for (uint i = 0; i < _contractsAddresses.length; i++) {
             contractsWhiteListMap[_contractsAddresses[i]] = _contractsWhitelisted[i];
-            emit UpdatedContractsWhitelist(_contractsAddresses[i], _contractsWhitelisted[i]);
+            emit UpdatedContractsWhitelist(_contractsAddresses, _contractsWhitelisted);
```
orenyodfat, line 85: this should be out of the loop :)

contracts/schemes/DxDaoSchemeConstraints.sol
```diff
@@ -101,7 +105,7 @@ contract DxDaoSchemeConstraints is SchemeConstraints {
         "invalid length _tokensPeriodLimits");
         for (uint i = 0; i < _tokensAddresses.length; i++) {
             periodLimitToken[_tokensAddresses[i]] = _tokensPeriodLimits[i];
-            emit UpdatedPeriodLimitsTokens(_tokensAddresses[i], _tokensPeriodLimits[i]);
+            emit UpdatedPeriodLimitsTokens(_tokensAddresses, _tokensPeriodLimits);
```
orenyodfat, line 104: this should be out of the loop.
you should have a test which check that only a single event is emitted here

test/genericschememulticall.js
```diff
@@ -264,45 +264,30 @@ contract('GenericSchemeMultiCall', function(accounts) {
...
        try {
-          await testSetup.genericSchemeMultiCall.execute(proposalId);
+          await testSetup.GenericSchemeMultiCall.execute(proposalId);
           assert(false, "cannot transfer eth amount");
         } catch(error) {
           helpers.assertVMException(error);
         }
     });
 
-    it("schemeconstrains token value exceed limit", async function() {
```
orenyodfat, line 281: Why this test was removed ?

test/genericschememulticall.js
```diff
@@ -264,45 +264,30 @@ contract('GenericSchemeMultiCall', function(accounts) {
...
-     });
-
     it("execute proposeVote -negative decision - check action - with GenesisProtocol", async function() {
        var actionMock =await ActionMock.new();
        var standardTokenMock = await ERC20Mock.new(accounts[0],1000);
        var testSetup = await setup(accounts,[actionMock.address],0,true,standardTokenMock.address);
 
        var callData = await createCallToActionMock(testSetup.org.avatar.address,actionMock);
-       var tx = await testSetup.genericSchemeMultiCall.proposeCalls([actionMock.address],[callData],[0],helpers.NULL_HASH);
+       var tx = await testSetup.GenericSchemeMultiCall.proposeCalls([actionMock.address],[callData],[0],helpers.NULL_HASH);
```
orenyodfat, line 287: Why change that to UpperCase .?

test/genericschememulticall.js
```diff
@@ -49,7 +49,7 @@ const setup = async function (accounts,
                               useSchemeConstraint = true) {
    var testSetup = new helpers.TestSetup();
    testSetup.standardTokenMock = await ERC20Mock.new(accounts[1],100);
-   testSetup.genericSchemeMultiCall = await GenericSchemeMultiCall.new();
+   testSetup.GenericSchemeMultiCall = await GenericSchemeMultiCall.new();
```
orenyodfat, line 52: why rename that ?

test/genericschememulticall.js
```diff
@@ -427,6 +431,83 @@ contract('GenericSchemeMultiCall', function(accounts) {
...
+        helpers.assertVMException(error);
+      }
+    });
+
+    it("cannot initialize contraints with invalid array length", async function() {
+      var dxDaoSchemeConstraintsArray=await DxDaoSchemeConstraints.new();
+      try {
+        await dxDaoSchemeConstraintsArray.initialize(
+          accounts[0],
+          0,
```
orenyodfat, line 456: this will revert due to 0 periodSize and not due to wrong array length.

test/genericschememulticall.js
```diff
@@ -445,6 +526,23 @@ contract('GenericSchemeMultiCall', function(accounts) {
...
 
+    it("execute proposeVote with multiple calls with votingMachine without whitelisted spender", async function() {
+      var actionMock =await ActionMock.new();
+      var standardTokenMock = await ERC20Mock.new(accounts[0],1000);
+      var testSetup = await setup(accounts,[actionMock.address],0,true,standardTokenMock.address);
+      var encodedTokenApproval= await createCallToTokenApproval(standardTokenMock, accounts[3], 1000);
+      var callData1 = await createCallToActionMock(testSetup.org.avatar.address,actionMock);
+      try {
+         await testSetup.genericSchemeMultiCall.proposeCalls(
+           [actionMock.address],
```
orenyodfat, line 537: this will revert due to wrong array length and not due to the spender not white listed

PR #803, opened 2020-11-09
add SimpleSchemeConstraint.sol

orenyodfat:
- Add descriptionHash to SchemeConstraint.sol interface.
   This can be use to give more details description on the constraints which can be used by the UI., 
    (e.g IPFS file ..which give info re white list contracts names, Abis..) 
- Add SimpleSchemeConstraint.sol with only white list contract contraints.


ben-kaufman (approved):
The code looks good to me, just need to fix coverage

PR #804, opened 2020-11-11
Multi Call factory

ben-kaufman:


contracts/schemes/GenericSchemeMultiCallFactory.sol
```diff
@@ -0,0 +1,49 @@
...
+        address simpleSchemeConstraints;
+        if (_contractsWhiteList.length > 0) {
+            simpleSchemeConstraints = new SimpleSchemeConstraints();
+            SimpleSchemeConstraints(simpleSchemeConstraints).initialize(_contractsWhiteList, _descriptionHash);
+        }
+        
+        bytes32 voteParams;
+        if (_voteParamsType == FAST) {
+            // Fast params hash
+            voteParams = bytes32(0x1b46f925b15bc0590168247d8df7f72773ca64dec3334183b5387dd3945f7f2e);
```
orenyodfat, line 33: I think that we might need to check here first if the params are set in the voting machine..
if not it should set it up.

contracts/schemes/GenericSchemeMultiCallFactory.sol
```diff
@@ -0,0 +1,49 @@
...
+    uint8 public constant CUSTOM = 0;
+    uint8 public constant FAST = 1;
+    uint8 public constant NORMAL = 2;
+    uint8 public constant SLOW = 3;
+
+    function createGenericSchemeMultiCallSimple(
+        Avatar _avatar,
+        IntVoteInterface _votingMachine,
+        uint8 _voteParamsType,
+        bytes32 _voteParamsHash,
```
orenyodfat, line 19: lets add also the gov params..
something like https://github.com/daostack/arc/blob/master-2/contracts/schemes/ArcScheme.sol#L33

orenyodfat (changes requested):

orenyodfat (changes requested):
lets move it under contracts/utils/

test/genericschememulticallfactory.js
```diff
@@ -0,0 +1,113 @@
...
+
+const setup = async function () {
+   var testSetup = new helpers.TestSetup();
+   testSetup.genericSchemeMultiCallFactory = await GenericSchemeMultiCallFactory.new();
+   return testSetup;
+};
+
+contract('genericSchemeMultiCallFactory', function(accounts) {
+  before(function() {
+    helpers.etherForEveryone(accounts);
```
orenyodfat, line 69: I think this is redundant

contracts/utils/GenericSchemeMultiCallFactory.sol
```diff
@@ -0,0 +1,96 @@
...
+
+    function createGenericSchemeMultiCallSimple(
+        Avatar _avatar,
+        IntVoteInterface _votingMachine,
+        uint8 _voteParamsType,
+        uint256[11] memory _votingParams,
+        address _voteOnBehalf,
+        address[] memory _contractsWhiteList,
+        string memory _descriptionHash
+    ) public returns(address) {
```
orenyodfat, line 24: `public`  -> `external`

contracts/utils/GenericSchemeMultiCallFactory.sol
```diff
@@ -0,0 +1,96 @@
...
+        (uint256 queuedVoteRequiredPercentage, , , , , , , , , , , ,) =
+        genesisProtocol.parameters(voteParamsHash);
+        if (queuedVoteRequiredPercentage == 0) {
+            //params not set already
+            genesisProtocol.setParameters(voteParams, _voteOnBehalf);
+        }
+        genericSchemeMultiCall.initialize(
+            _avatar, _votingMachine, voteParamsHash, SchemeConstraints(simpleSchemeConstraints)
+        );
+        return address(genericSchemeMultiCall);
```
orenyodfat, line 94: lets emit an event with the address...

contracts/utils/GenericSchemeMultiCallFactory.sol
ben-kaufman: I can't this gives a stack too deep error. But it doesn't really matter anyway...

orenyodfat (approved):

PR #805, opened 2020-11-12
only register scheme can propose and be voted

orenyodfat:


ben-kaufman (approved):
Need to fix coverage...

PR #815, opened 2021-01-28
Added DxDaoGenericSchemeMultiCall

nicoelzer:
- Adds a specific implementation of GenericSchemeMulticall that prevents sending ETH in proposals


orenyodfat:
might be that the schemeConstraint is a better place to add such constraint.
see #816

nicoelzer:
Closed as of duplicate #816

